PR #6, opened 2017-12-13
Added wget flag to reset connection

Manouchehri:
Since install-sh.sh is a part of the install process, we probably don't want to wait for the long default of 15 minutes (https://github.com/chapmanb/bcbio-nextgen/issues/1487).

Was having issues on my build box before applying this change (connections to `*.s3.amazonaws.com` kept hanging).

### DIFF
--- a/cmake/install-share.sh
+++ b/cmake/install-share.sh
@@ -64,7 +64,7 @@ mkdir -p "$SUPPORT_DIR"
 # Get archive using wget.
 ARCH_URL="https://github.com/avast-tl/retdec-support/releases/download/$VERSION/$ARCH_NAME"
 echo "Downloading archive from $ARCH_URL ..."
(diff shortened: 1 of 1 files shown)


s3rvac:
Thank you for the suggestion. I agree that setting `--read-timeout` to a lower-than-default value may be desirable. However, do we really want it to be set to a single second? Shouldn't we set it to, say, 10 seconds? Otherwise, if there is ever a one-second read delay during the download process, the entire download gets restarted. That seems pretty harsh to me :-).

Manouchehri:
Done.

s3rvac:
Thank you. Could you please squash all the commits together and base the resulting commit on the current `origin/master`?

Manouchehri:
Done!

s3rvac:
Thanks!

PR #17, opened 2017-12-13
macOS build fixes and instructions

### LINKED ISSUE #7: Add support for macOS build

stek29:
Fixes #7 

See avast-tl/llvm#1 and avast-tl/fileformat#2


Manouchehri:
Nice! How large is your output folder? Maybe we can make a Homebrew package? (They handle pushing to Bintray too.)

stek29:
@Manouchehri 3.9 GB :S

mewmew:
> @Manouchehri 3.9 GB :S

Most likely this is due to the yara_patterns in share. Mine is ~3.7 GB.

```
u@x220 /u/s/generic> du -hs yara_patterns/
3.8G	yara_patterns/
```

petmac:
Thanks for working on this. I had a cursory look but didn't go further than that.

There seemed to be a bunch of `CMakeLists.txt` files that checked for Clang using `... STREQUAL "Clang"` which obviously won't detect `AppleClang`. Changing `STREQUAL` to `MATCHES` should fix those.

In your PR, looks like `README.md` was deleted/renamed?

stek29:
@petmac there's only one file checking for compiler, and it should not confuse AppleClang and Clang there :)

stek29:
@petmac Wow, thanks, I've cleared README.md somehow :\

ostefano:
@stek29 I've tried to compile your branch on a 10.13.2 system. It complains about a missing library (which is not available on osx afaik). Maybe some makefiles with a missing switch?

```
Scanning dependencies of target capstone2llvmirtool
[ 17%] Building CXX object deps/capstone2llvmir/src/capstone2llvmirtool/CMakeFiles/capstone2llvmirtool.dir/capstone2llvmir.cpp.o
[ 17%] Linking CXX executable capstone2llvmir
ld: library not found for -lrt
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [deps/capstone2llvmir/src/capstone2llvmirtool/capstone2llvmir] Error 1
make[1]: *** [deps/capstone2llvmir/src/capstone2llvmirtool/CMakeFiles/capstone2llvmirtool.dir/all] Error 2
make: *** [all] Error 2
```

stek29:
@ostefano 
> See avast-tl/llvm#1 and avast-tl/fileformat#2

sarnau:
@stek29 @petmac is correct with the STREQUAL vs MATCHES comment. Without that you do get an error "Unsupported compiler. Supported: GCC, Clang." in the openssl dependency.

stek29:
@sarnau Are you sure you have fileformat and llvm with my patches merged?

[See this](https://github.com/avast-tl/fileformat/blob/32b0c72052d78dbf41dec9b1ebf8cbed090e168c/deps/openssl/CMakeLists.txt#L40-L44)

Just making MATCHES is not a good idea. If it prints "Supported: GCC, Clang", then it thinks it's Linux, and that's not correct. So better it says "Unsupported compiler" than silently fails with lots of issues about trying to use linux kernel crypto features on macOS.

newkit:
Just out of curiousity - how did you avoid that the original Apple `flex`version got picked up? My build was failing until I renamed the original Apple files.

stek29:
@newkit README is made for reading it.

newkit:
Oh really? ;-) I know it is mentioned there and I had installed it through homebrew. And changed the `PATH` variables as mentioned, but still `yaramod`picked up Apples `/usr/bin/flex`and failed building.

stek29:
@newkit Did you also change two other variables? Did you put homebrew's flex/bison in the beginning of PATH?

PR #21, opened 2017-12-14
fix path detection for symlinks & support cygwin paths via cygpath

apfeltee:
Right now, the scripts used by retdec (decompile.sh, unpack.sh, etc) will fail if they're symlinked, i.e., `decompile.sh` symlinked to `~/retdec-decompile`, because both `${BASH_SOURCE}` and `$0` will always point at the symlinked path, rather than the original, so `$SCRIPTPATH` would become invalid.

This is easy enough to fix, by doing `dirname "$(readlink -e "$0")"`, rather than the pretty ancient style of `cd`ing to `dirname "$0"` (pretty sure this has been out of date since the inception of Linux itself ...).  
The other problem is that the scripts will also fail on Cygwin, due to cygwin's virtual paths (`c:` becomes `/cygdrive/c/`, and so on), which are *not* transformed by readlink. Instead, one needs to use cygpath, to get the real windows path. So I've added `get_realpath` to `utils.sh`, to check if the scripts are running on Cygwin, and if they do, to use `cygpath -ma` (`-a` = absolute path, and `-m` = mixed, i.e., forward slashes rather than backward slashes). Otherwise it just uses `readlink`.

Maybe this isn't the best way, but it makes retdec usable on Cygwin, and especially for anyone symlinking the scripts.

### DIFF
--- a/scripts/config.sh
+++ b/scripts/config.sh
@@ -3,7 +3,7 @@
 # Decompiler's configuration. This file should be included in every decompilation script.
 #
 
-SCRIPTPATH="$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )"
+SCRIPTPATH="$(dirname "$(readlink -e "$0")")"
 
 if [ -z "$DECOMPILER_CONFIG" ]; then
 	DECOMPILER_CONFIG="./config.sh"
--- a/scripts/decompile-all.sh
+++ b/scripts/decompile-all.sh
@@ -8,7 +8,7 @@
 #  - the `find` and `timeout` commands
 #
 
-SCRIPTPATH="$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )"
+SCRIPTPATH="$(dirname "$(readlink -e "$0")")"
 
 if [ -z "$DECOMPILER_UTILS" ]; then
 	DECOMPILER_UTILS="$SCRIPTPATH/utils.sh"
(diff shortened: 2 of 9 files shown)


s3rvac:
Thank you. This definitely makes sense to support. I will look into the provided patch very soon. In the meantime, could I ask you to:
1. Fix the indentation in `get_realpath()` (we use tabs in scripts).
2. Add a comment before the function (as in, e.g., `print_warning()`).
3. Add a comment into the function describing why `cygpath` has to be used on CYGWIN.

apfeltee:
hi @s3rvac,

Done and done! I'm very sorry about the delay, by the way.

s3rvac:
Thank you, I will check the changes. BTW, I just noticed that you have committed both commits under `John Doe <root@localhost>`. Is this on purpose, or are you just missing your real name and email in your Git configuration file?

If you want, you can squash the two commits into one and fix the name and email.

s3rvac:
Since there was no activity in the last 30 days and the PR cannot be merged without many conflicts, I took the liberty of incorporating your work into the repository by myself. I also had some minor comments concerning the PR, which I reflected in the commits (d0086e63c6cc35ca55c3527c2ccd970f2c659f10, 90767ab6f808601bcf42697c6101bfc2bd4a43de). I hope that you do not mind. Your fixes were good, so it would be a shame not to incorporate them. Thank you for them!

Issue #19, opened 2017-12-13
Build failed: cannot find -ltinfo


ner0x652:
Hi. This is the error message during a build on a 
`Linux machine 4.10.0-42-generic #46~16.04.1-Ubuntu SMP Mon Dec 4 15:57:59 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux` :

`
/usr/bin/ld: cannot find -ltinfo
/usr/bin/ld: cannot find -ltinfo
collect2: error: ld returned 1 exit status
deps/capstone2llvmir/src/capstone2llvmirtool/CMakeFiles/capstone2llvmirtool.dir/build.make:169: recipe for target 'deps/capstone2llvmir/src/capstone2llvmirtool/capstone2llvmir' failed
make[2]: *** [deps/capstone2llvmir/src/capstone2llvmirtool/capstone2llvmir] Error 1
CMakeFiles/Makefile2:1515: recipe for target 'deps/capstone2llvmir/src/capstone2llvmirtool/CMakeFiles/capstone2llvmirtool.dir/all' failed
make[1]: *** [deps/capstone2llvmir/src/capstone2llvmirtool/CMakeFiles/capstone2llvmirtool.dir/all] Error 2
Makefile:151: recipe for target 'all' failed
make: *** [all] Error 2
`

TcM1911:
For Ubuntu, ensure libtool autoconf libtinfo-dev zlib1g-dev flex bison are installed

s3rvac:
Yes, the error is caused by missing [`libtinfo`](https://packages.ubuntu.com/search?keywords=libtinfo-dev). We have updated the [build instructions](https://github.com/avast-tl/retdec#build-and-installation). Can you please try to install all the required packages and retry the build?

ner0x652:
Yes, I'll try. Thank you for the help.

tpruzina:
Pretty sure that some distros ship ncurses without tinfo (ubuntu being one of them). Gentoo ships tinfo support optionally. I kinda wanted to report this before but then I thought that it was one of your dependencies that relied on this rather than retdec.

liushuyu:
Hi,

I have tried symlink `libncurses.so` to `libtinfo.so` and it passed the compilation and the resulting binary seemed working. Another approach is to use the patch here: 
[0001-tinfo-to-ncurses.zip](https://github.com/avast-tl/retdec/files/1561568/0001-tinfo-to-ncurses.zip)
[...]

PR #22, opened 2017-12-14
Added apt-get packages

Manouchehri:
Solves #19.


s3rvac:
Good idea! Thank you :heart:

PR #37, opened 2017-12-15
Add Fedora dependency install command

m3gat0nn4ge:


metthal:
Thank you, but I already had the list of all required packages for Fedora prepared to be committed when you created the pull request.

Some notes:
1. You don't need `openssl-devel`. We download OpenSSL using `ExternalProject`, so you don't have to have OpenSSL headers installed. We also discourage from using system OpenSSL (yet, we might possibly want to change this in the future).
2. You omitted a lot of packages. Even if they are present in the default installation of Fedora, I personally know many Linux users who cleanup their systems and uninstall all packages they don't use. Users must be able to install all dependencies by copy-pasting install command from our README.

Thank you anyways, we appreciate any help.

m3gat0nn4ge:
Strange.. it didn't find openssl during make, that was why I added it to the dependency list..

s3rvac:
Yes, this is indeed strange, as none of our dependencies require a system version of OpenSSL.

PR #53, opened 2017-12-19
[WIP] Add Travis CI

lilianmoraru:
The idea in the end is to create Linux binaries with the help of Travis(To reduce expectations: I don't yet plan to do this task), but not yet available here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,81 @@
+language: cpp
+cache:
+  apt:    true
+  ccache: true
+
+# Use https (public access) instead of git for git-submodules. This modifies only Travis-CI behavior!
+# disable the default submodule logic
+git:
+  submodules: false
+
+# use sed to replace the SSH URL with the public URL, then init and update submodules
+before_install:
+  - sed -i 's/git@github.com:/https:\/\/github.com\//' .gitmodules
+  - git submodule update --init --recursive
+
+addons:
+  apt:
+    packages:
+    - build-essential
+    - cmake
+    - perl
+    - python3
+    - bash
+    - coreutils
+    - wget
+    - bc
+    - doxygen
+    - graphviz
+    - upx
+    - flex
+    - bison
+    - zlib1g-dev
+    - libtinfo-dev
+    - autoconf
+    - automake
+    - pkg-config
+    - m4
+    - libtool
+
+matrix:
+  fast_finish: true
+  include:
+    - os: linux
+      compiler: gcc-7
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - g++-7
+      env:
+        - CMAKE_FLAGS="-DCMAKE_BUILD_TYPE=Debug
+                       -DCMAKE_C_COMPILER=gcc-7
+                       -DCMAKE_CXX_COMPILER=g++-7"
+    - os: linux
+      compiler: clang
+      env: CMAKE_FLAGS="-DCMAKE_BUILD_TYPE=Debug
+                        -DCMAKE_C_COMPILER=clang
+                        -DCMAKE_CXX_COMPILER=clang++"
+  allow_failures:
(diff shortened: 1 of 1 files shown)


lilianmoraru:
You can follow the status of the build here: https://travis-ci.org/lilianmoraru/retdec/builds/318839555

lilianmoraru:
Travis overwrites CC and CXX, need to handle that...

lilianmoraru:
Seems like `capstone2llvmir` won't be compiling on clang...: https://travis-ci.org/lilianmoraru/retdec/jobs/318848946#L1045

lilianmoraru:
Timed out - takes to much time to compile all the dependencies.
I will make it compile only `capstone2llvmir` - let it cache it and then you can switch back to building everything.

s3rvac:
Thank you for the effort. A couple of notes from me:
* I am afraid we would not be able to use the free version of Travis CI due to their time limits. Nevertheless, its worth a shot as I might be wrong.
* As for the unsuccessful Clang build, it failed with the following error:
    ```
    tl-cpputils/range.h:179:39: error: no template named 'enable_if_t' in namespace 'std'; did you mean 'enable_if'?
    ```
    RetDec build requires a C++14-capable compiler and standard library implementation. [`std::enable_if_t`](http://en.cppreference.com/w/cpp/types/enable_if) was introduced into the standard library in C++14, but it seems that the version of the standard library that was used during the build did not yet have that implementation.

lilianmoraru:
@s3rvac You could use Travis CI if you roll it incrementally initially, with ccache.
The `capstone2llvmir` target took _~30 minutes_ to build the first time, now with ccache it took only _74 seconds_(if you switch from the default linker, you will get this under 1 minute).

The steps here are:
- Set Travis on your repo, before accepting this PR
- When the PR is merged, let it make a full build with the `capstone2llvmir`, this will cache the most time-consuming build(LLVM) and then you can remove the line with the `capstone2llvmir` build and uncomment the other 2, to run the full build - let it finish and cache the build.
After this, you will have free time to even run some tests.

> RetDec build requires a C++14-capable compiler

You have `set(CMAKE_CXX_STANDARD 14)` in the CMakeLists.txt and Clang 5 fully supports it.
It seems to me that the issue here is that it uses the header from GCC 4.8, installed in the system.
Have to see yet what is the problem here - I also don't like that the GCC build ends up using GCC 4.8 through ccache, for some of the components.

s3rvac:
> You could use Travis CI if you roll it incrementally initially, with ccache.

Unfortunately, I do not think that a partial support that requires manual "hacking" in `.travis.yml` will be desirable. We are currently discussing which CI system we should use, and we will definitely want to go with one that will enable hassle-free setup, use, and will be able to build and test the entire repository.

> You have set(CMAKE_CXX_STANDARD 14) in the CMakeLists.txt and Clang 5 fully supports it.
> It seems to me that the issue here is that it uses the header from GCC 4.8, installed in the system.
> Have to see yet what is the problem here - I also don't like that the GCC build ends up using GCC 4.8 through ccache, for some of the components.

Yeah, GCC 4.8 is not supported due to lacking C++14 support in the standard library. RetDec requires at least GCC 4.9.

lilianmoraru:
> Unfortunately, I do not think that a partial support that requires manual "hacking" in .travis.yml will be desirable.

It is not "partial support" or "manual hacking", it is just a first step to cache the LLVM build and then it is "full support", "no manual hacking".
If you will change files in LLVM, it will rebuild the object files that need to be modified - it is still a full build.

s3rvac:
I understand. I did not mean to discourage you. I am just afraid of the problems that such an incremental solution might cause us. For example, if we ever wanted to clear the cache and start building from scratch (e.g. if the incremental build starts failing), we would need to do the same thing again (build LLVM, then possibly build something else, then a full build). Otherwise, the build will start timeouting. Or is there something I am missing?

PR #65, opened 2017-12-21
expand citability by adding CITATION file

katrinleinweber:
Hello!

This is a copy your 2017 entry on https://retdec.com/publications/ incorporated into the repo as suggested in https://www.software.ac.uk/blog/2013-09-02-encouraging-citation-software-introducing-citation-files. Adding it here may make it simpler for people to cite your software. How about it?

Also, it would be possible edit this PR to:

- [ ] use `@software{...`
- [ ] add a DOI through https://guides.github.com/activities/citable-code/
- [x] include a `url = {...}`

Cheers!

### DIFF
--- a/CITATION
+++ b/CITATION
@@ -1,0 +1,8 @@
+@Misc{RETDEC_BOTCONF_17,
+    author = {J. K\v{r}oustek and P. Matula and P. Zemek},
+    title = {RetDec: An Open-Source Machine-Code Decompiler},
+    year = {2017},
+    month = {December},
+    howpublished = {[talk]},
+    note = {Presented at Botconf 2017, Montpellier, FR}
+}
--- a/CITATION
+++ b/CITATION
@@ -1,4 +1,4 @@
-@Software{RetDec,
+@Misc{RetDec,
     author = {{Avast Software}},
     title = {{RetDec}: A Retargetable Machine-Code Decompiler},
     howpublished = {\url{https://retdec.com/}}
--- a/CITATION
+++ b/CITATION
@@ -1,8 +1,5 @@
-@Misc{RETDEC_BOTCONF_17,
-    author = {J. K\v{r}oustek and P. Matula and P. Zemek},
-    title = {RetDec: An Open-Source Machine-Code Decompiler},
-    year = {2017},
-    month = {December},
-    howpublished = {[talk]},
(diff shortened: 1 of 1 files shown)


s3rvac:
Thank you for the idea! We definitely want to do this, but we are not yet sure what to include into the `CITATION` file. We will need to discuss this in our team. Let me get back to you during the first week of January.

s3rvac:
We have decided to go with the following BibTeX entry:
```
@Misc{RetDec,
    author = {{Avast Software}},
    title = {{RetDec}: A Retargetable Machine-Code Decompiler},
    howpublished = {\url{https://retdec.com/}}
}
```
Could you please update the PR to reflect this change?

katrinleinweber:
Will do :-) If you want to focus on the software, how about the `@software{…}` citation type, instead of `Misc`? I initially choose the latter only because I wasn't sure how to categorise a talk in BibTeX.

s3rvac:
AFAIK `@Software` is only available if you use the [BibLaTeX](https://ctan.org/pkg/biblatex) package, which is why we decided to go with `@Misc`.

CITATION
```diff
@@ -0,0 +1,5 @@
+@Software{RetDec,
```
s3rvac: I think there was a misunderstanding. We would like to use `@Misc` here as `@Software` is only available if you use the [BibLaTeX](https://ctan.org/pkg/biblatex) package.

katrinleinweber:
Argh, sorry! I had the variant with `@software` prepared on my main computer, then read your reply on another computer, but still committed the first variant.

katrinleinweber:
Can you squash in GitHub's web interface? Or should I really do that offline & force-push the new commit here? Or into a new PR?

s3rvac:
Thank you ❤️ I will squash them from here.

katrinleinweber:
I'm keeping eyes and ears open for wider supported for the citation type `@software`.

s3rvac:
Yes, this, and also the fact that when you try to use `@Software{XYZ` with standard LaTeX+BibTeX, you get
```
Warning--entry type for "XYZ" isn't style-file defined
```
Consequently, `@Misc` is more portable, as you do not need any additional package or non-standard bibliography style.

PR #75, opened 2017-12-31
added Mac OS specific sha256sum variant in install-share.sh

newkit:
There is no sha256sum on Mac OS. In order to make install-share.sh work,
the command was changed in order to use shasum -a 256 on Mac OS.

### DIFF
--- a/cmake/install-share.sh
+++ b/cmake/install-share.sh
@@ -14,6 +14,7 @@ fi
 VERSION_FILE_NAME="version.txt"
 ARCH_SUFFIX="tar.xz"
 
+SHACMD='sha256sum'
 SHA256SUM_REF="4ecbe72d38d6e95f057c5d18c79e25b5340265416687368ce7e5fd1ebb744b32"
 VERSION="2017-12-15"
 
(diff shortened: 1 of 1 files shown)


newkit:
Minor change in order to support Mac OS.

stek29:
Heh, I haven't noticed this because I had GNU coreutils in my path and they have sha256sum

s3rvac:
Thank you for the PR. On both Linux and Windows, we require [GNU coreutils](https://www.gnu.org/software/coreutils/coreutils.html) to be installed. I wonder if we should also require `coreutils` on macOS to make the build instructions more consistent. What do you think?

newkit:
Yeah, installing `coreutils` is maybe the more elegant solution, so adding them as a build requirement for Mac OS should be implemented rather then my PR. Add this change of the `PATH` variable in order to include `coreutils`:

```
export PATH="/usr/local/opt/coreutils/libexec/gnubin:$PATH"
```

s3rvac:
Alright, I will add `coreutils` into the list of required packages.

s3rvac:
Added in cbd7685d31185a0d781028cfce24574feed07fc2. Thank you for noticing that `coreutils` are missing in our instructions!

PR #98, opened 2018-01-21
Use `gsha256sum` instead of `sha256sum` if it's available.

comex:
GNU coreutils are sometimes installed prefixed with 'g', including on macOS with Homebrew and on BSDs.  The README was recently changed to suggest adding `$(brew --prefix coreutils)/libexec/gnubin` to $PATH (see #75), but this way is better as it works with the default PATH.

Also remove that bit from the README.

(This still requires coreutils to be installed through Homebrew.  #75 originally proposed to use a different utility to compute SHA256 hashes on macOS, but @s3rvac expressed a desire to standardize on coreutils.  Makes sense, but it's still good to minimize the number of manual steps required to successfully build, so I thought this was a reasonable compromise.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,7 +153,7 @@ Packages should be preferably installed via [Homebrew](https://brew.sh).
 * [Bison](https://www.gnu.org/software/bison/) (version >= 3.0)
 * [Flex](https://www.gnu.org/software/flex/) (version >= 2.6)
 * [autotools](https://en.wikipedia.org/wiki/GNU_Build_System) ([autoconf](https://www.gnu.org/software/autoconf/autoconf.html), [automake](https://www.gnu.org/software/automake/), and [libtool](https://www.gnu.org/software/libtool/))
-* [coreutils](https://www.gnu.org/software/coreutils) (ensure that you have `$(brew --prefix coreutils)/libexec/gnubin` in your `PATH`)
+* [coreutils](https://www.gnu.org/software/coreutils)
 * [wget](https://www.gnu.org/software/wget/)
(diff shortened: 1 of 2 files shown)


s3rvac:
Thank you for the PR. I definitely agree that we should minimize the number of manual steps required for a successful build and run. With this in mind, the PR is definitely a step forward. However, I am not sure if `sha256sum` is the only coreutils command that we use. For example, in the decompilation scripts, we also use `readlink`, and possibly some other coreutils commands as well. Are you sure that decompilations can be run without `$(brew --prefix coreutils)/libexec/gnubin` in `PATH`? Maybe we should add this requirement into the [Run on macOS](https://github.com/avast-tl/retdec#macos) section or, better, use a similar approach to the one that you used. Let me know what you think.

s3rvac:
In commit 0f1281e536bfe2e6053f702138114de51066335d, I have updated the installation script to try the following three commands and use the first one that is available: `gsha256sum`, `shasum`, and `sha256sum`. So, in order to _install_ RetDec, GNU coreutils are no longer required. However, GNU coreutils are still needed to _run_ RetDec. So, at least, the installation will be easier. I know that I proposed to standardize on GNU coreutils in #75, but since `sha256sum` was the only required command, I think it makes sense support the native version as well.

In commit 9b2c10ee96b9664cd91ba1207a2b64ca88ecf9e4, I have updated all the scripts that used `readlink` to try `greadlink` first. This simplifies the setting of `PATH` as GNU coreutils no longer need to appear before the built-in tools (GNU coreutils just need to be available in `PATH`).

All in all, from what I have observed, there are not many places where GNU coreutils would be required. So, I believe we might be able to get rid of them and use the default versions that ship with macOS. I will happily take a PR for that.

PR #102, opened 2018-01-23
Let realpath allocate a buffer itself

yangfl:
Forwarded from https://bugs.debian.org/888067

`realpath` has the ability to allocate buffer itself. This could avoid depending on fix-sized buffer.

### DIFF
--- a/src/utils/filesystem_path.cpp
+++ b/src/utils/filesystem_path.cpp
@@ -168,9 +168,15 @@ public:
 
 	virtual std::string getAbsolutePath() override
 	{
+#ifdef PATH_MAX
 		char absolutePath[PATH_MAX] = { '\0' };
 		if (realpath(_path.c_str(), absolutePath) == nullptr)
 			return {};
+#else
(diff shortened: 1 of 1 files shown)


metthal:
Thank you for PR! If there are problems with `PATH_MAX` constant in some compilers, I can see merging this PR although it uses unnecessary heap allocation. However, right now it is missing `free()` call because as stated in man page for `realpath`:

```
If  resolved_path  is  specified as NULL, then realpath() uses malloc(3) to allocate a buffer of up to PATH_MAX bytes to hold the resolved pathname, and returns a pointer to this buffer. The caller should deallocate this buffer using free(3).
```

yangfl:
@metthal Hi, I've added a compiler macro that will prefer stack allocation when `PATH_MAX` available. Please review them.

metthal:
I've merged your changes. I also added the memory freeing into `SCOPE_EXIT` in case of exception during `std::string` allocation. Thank you for your contribution.

Issue #258, opened 2018-03-29
macOS: retdec-* breaks due to explicit use of #!/bin/bash


AshKetch-um:
On macOS the built-in `bash` and the associated tools are incompatible with many different things being done by `retdec-*` scripts. The solution is for this is to use homebrew or Mac Ports to provide a GNU compatible environment (see: [Run on MacOS](https://github.com/avast-tl/retdec#macos)).

There are a number of complicating factors between homebrew/ports and GNU-targeted tools however, one of the key issues is the collisions between built-in tools (e.g. `bash`, `sed`, `redline`) and those provided by these package managers. A common solution is to place the path to the GNU tools symlink tree earlier in the PATH than the built-in tools (e.g. `export PATH=/opt/local/bin:/opt/local/sbin:/opt/local/libexec/gnubin`) thus allowing GNU targeted tools to work without having to be specifically aware of the quirks of macOS/brew/ports.

Most of these issues have been identified and resolved as part of the `README.md` however, the `retdec-*` scripts explicitly utilize `#!/bin/bash` which will cause them to execute using the built-in `bash` rather than that of the current shell. This results in `retdec-*` only working when run using a dedicated brew/ports `bash` instance (not the current shell) and when run from the installation directory of retdec. This is because the explicit `#!/bin/bash` is causing the built-in `bash` to be used in place of the currently running shell, causing cascading issues in a number of different places (see: Observed Behavior below).

The fix that worked for me is replacing the explicit usage of `#!/bin/bash` in all to the shell scripts with the more environmentally friendly version `#!/usr/bin/env bash`. This allows the running shell (e.g. `/opt/local/bin/bash` to be used and thus, allows retdec's installation path to included into the PATH and thus run from anywhere. ( Pull Request #259 )

**Environment**
macOS 10.13.3
Bash: 4.4.19(1)-release (x86_64-apple-darwin17.4.0)
Mac Ports
changeset: 31cf35fff9dce8de081a8705653f2e56430ce270

**Steps to Reproduce**
Download, Build and Install on MacOS
```
sudo port selfupdate
# See #257
sudo port install cmake git perl5 python27 bison flex autoconf automake libtool coreutils wget gsed doxygen upx graphviz

# See #98
export PATH=/opt/local/bin:/opt/local/sbin:/opt/local/libexec/gnubin:$PATH

git clone https://github.com/avast-tl/retdec
cd retdec
mkdir build && cd build
[...]

PR #259, opened 2018-03-29
Issue #258: Replace usage of explicit '#!/bin/bash' with more environmentally friendly '#!/usr/bin/env bash'

AshKetch-um:
Issue #258


s3rvac:
Thank you!

Issue #270, opened 2018-04-04
Add support for limiting overall memory when running decompilations and tools


s3rvac:
We should add support for limiting the overall memory when running decompilations and tools to prevent issues like #36 and #81 (_black screen when decompiling on Windows_). On Linux, it is pretty easy to limit virtual memory via `ulimit -v`, but on Windows, this is not that straightforward.

Our scripts and tools should have a parameter by which one could control the maximal virtual memory (e.g. `--max-memory`). By default, this should be set to half of the available memory on the system. When a tool has allocated more than this limit, further allocations should fail with `std::bad_alloc` instead of starting swapping.

Note: We should of course try to reduce the memory requirements of our tools (see e.g. #13). This new parameter represents just a precaution to reduce swapping, and should be taken as such.

s3rvac:
Implemented in #272.

Details:
* Added a new module `memory` into `retdec::utils`, which provides ways of limiting the available memory of the current process (b63d814dfb6e094d571dcd54e80dc5a8b4b9d5a1). Supported platforms are Windows, Linux, and macOS.
* Added new parameters `--max-memory` and `--max-memory-half-ram` into `retdec-fileinfo` (bf0a1c64ab9021552705b4bc840b620b57d6e927). By default, there is no memory limit when you run the tool.
* Added new parameters `-max-memory` and `-max-memory-half-ram` into `retdec-bin2llvmir` and `retdec-llvmir2hll` (a7516cb698e8628853798370ce28c1a1885795ed and 349c5952760b23f270f1c88d31db26d5a43ae887). By default, there is no memory limit when you run the tools.
* Added new parameters `--max-memory` and `--no-memory-limit` into `retdec-decompiler.sh` (23434fa0bc7b36364da92340306c0f623c3770ad). By default, the script limits the memory of the above tools into half of system RAM to prevent potential black screens (especially on Windows). Use `--no-memory-limit` to remove this limit.
* Regression tests were added in the `issue-270-limiting-overall-memory` branch (merge: https://github.com/avast-tl/retdec-regression-tests/commit/0b9c04f7aa21ae52228e2a54872adb89922300ff).

PR #272, opened 2018-04-11
Add support for limiting overall memory when running decompilations and tools

s3rvac:
Adds support for limiting overall memory when running decompilations and tools (#270).

Details:
* Added a new module `memory` into `retdec::utils`, which provides ways of limiting the available memory of the current process. Supported platforms are Windows, Linux, and macOS.
* Added new parameters `--max-memory` and `--max-memory-half-ram` into `retdec-fileinfo`. By default, there is no memory limit when you run the tool.
* Added new parameters `-max-memory` and `-max-memory-half-ram` into `retdec-bin2llvmir` and `retdec-llvmir2hll`. By default, there is no memory limit when you run the tools.
* Added new parameters `--max-memory` and `--no-memory-limit` into `retdec-decompiler.sh` By default, the script limits the memory of the above tools into half of system RAM to prevent potential black screens (especially on Windows). Use `--no-memory-limit` to remove this limit.
* Regression tests are in the `issue-270-limiting-overall-memory` branch, which will have to be merged after this PR is merged.

### DIFF
--- a/include/retdec/utils/memory.h
+++ b/include/retdec/utils/memory.h
@@ -1,0 +1,22 @@
+/**
+* @file include/retdec/utils/memory.h
+* @brief Memory utilities.
+* @copyright (c) 2017 Avast Software, licensed under the MIT license
+*/
+
+#ifndef RETDEC_UTILS_MEMORY_H
+#define RETDEC_UTILS_MEMORY_H
+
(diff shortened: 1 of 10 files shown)


MerovingianByte:
Thank you for this. Much appreciated.

PR #300, opened 2018-05-17
OpenSSL build on ARM architecture

metthal:
This PR fixes build of OpenSSL on ARM architectures for Windows & Linux. It also adds support for build on unrecognized architectures for Linux (`generic32`, `generic64`). Can you please review the changes? Thanks.


metthal:
I agree, we should check if `OPENSSL_CONFIGURE_ARCH` defined so I've added it. I'll wait for AppVeyor and Travis and then merge it if there is nothing else to change.

PR #308, opened 2018-05-24
simplifies docker container usage 

ba0f3:


README.md
```diff
@@ -224,31 +224,15 @@ docker build -t retdec:dev . -f Dockerfile.dev
...
-
-To decompile a binary, create a container to upload the binary to:
-```
-docker create --name retdec_init retdec
-```
-
-Upload the binary (note the destination directory should be a directory with read/write permissions, such as `/home/retdec/`):
-```
-docker cp <file> retdec_init:/destination/path/of/binary
+Make sure the folder writable by retdec
```
s3rvac: I am wondering: Why is this step necessary? I was able to run a decompilation even without it (the produced files were created under the user that ran the container). Wouldn't it suffice for the directory to be readable and writeable for the current user?

README.md
```diff
@@ -224,31 +224,15 @@ docker build -t retdec:dev . -f Dockerfile.dev
...
 ```
-
-Commit the copied files into the container image:
+chmod -R 0666 /path/to/local/folder
```
s3rvac: Are the permissions correct? `0666` makes the directory non-accessible. To enter a directory, it must have the `x` bit set, and `0666` removes that bit. So, after a decompilation, you will be unable to read the results.

README.md
```diff
@@ -224,31 +224,15 @@ docker build -t retdec:dev . -f Dockerfile.dev
...
-
 Run the decompiler:
 ```
-docker run --name retdec retdec:initialized retdec-decompiler.sh /destination/path/of/binary
-```
-
-Copy output back to host:
-```
-docker cp retdec:/destination/path/of/binary.c /path/to/save/file
+docker run --rm -v /path/to/local/folder:/destinaton retdec retdec-decompiler.sh /destinaton/binary
```
s3rvac: Typo: `destinaton` -> `destination`

README.md
s3rvac: I see. Thank you for the explanation!

README.md
ba0f3: im not sure about this :| Its still accesible for me

s3rvac:
```
$ mkdir test
$ touch test/file.txt
$ chmod -R 0666 test
chmod: cannot access 'test/file.txt': Permission denied
$ cat test/file.txt
cat: test/file.txt: Permission denied
```
I suggest using `0777` instead of `0666`, and changing only permissions on the directory by removing the `-R` parameter.

ba0f3:
I merged new changes into one commit

s3rvac:
Perfect, thank you!

Issue #317, opened 2018-05-28
Use FilesystemPath::exists() instead of fileExists()


s3rvac:
In `include/retdec/utils/file_io.h`, there is the following function that checks whether a given file exists:
```cpp
bool fileExists(const std::string& file);
```
I suggest removing this function and replacing its uses with `FilesystemPath::exists()` so we do not have two ways of doing the same thing. `FilesystemPath` is the preferred way of manipulating file paths, anyway.

PR #320, opened 2018-06-03
Use FilesystemPath::exists() instead of fileExists() (#317)

TheCodez:
Fix https://github.com/avast-tl/retdec/issues/317 by replacing all `fileExists `method calls with `FilesystemPath::exists()`


s3rvac:
Nice, thank you!

PR #328, opened 2018-06-09
Fix comparison of vtables and comparison of vtable items.

Alexander-Alekseev:
There were identical sub-expressions to the left and to the right of the '==' operator. In other words, vtable item/vtable was compared with itself (that results to 'true') instead of comparison with other vtable item/vtable.


s3rvac:
Thank you!

PR #331, opened 2018-06-09
Fix possible memory leak in MachOFormat::symtabCommand() method.

Alexander-Alekseev:
The symtabCommand method was exited without releasing the 'symbolTable' pointer.


s3rvac:
Nice catch, thanks!

Issue #164, opened 2018-02-06
Bin2llvmir: leaking memory


PeterMatula:
Bin2llvmir is leaking memory while processing this input: [leak.zip](https://github.com/avast-tl/retdec/files/1700262/leak.zip).

E.g. (and many other places):
```
==5589== 1,024 bytes in 1 blocks are indirectly lost in loss record 1,303 of 1,459
==5589==    at 0x4C2D52F: operator new(unsigned long) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==5589==    by 0x47DEC6: retdec::dwarfparser::DwarfLineContainer::loadAndGetDie(Dwarf_Die_s*, unsigned int) (in /home/peter/retdec-github/retdec/build/install/bin/retdec-bin2llvmir)
==5589==    by 0x477AE4: retdec::dwarfparser::DwarfFile::loadDIE(Dwarf_Die_s*, retdec::dwarfparser::DwarfBaseElement*&, int) (in /home/peter/retdec-github/retdec/build/install/bin/retdec-bin2llvmir)
==5589==    by 0x478167: retdec::dwarfparser::DwarfFile::loadCUtree(Dwarf_Die_s*, retdec::dwarfparser::DwarfBaseElement*, int) (in /home/peter/retdec-github/retdec/build/install/bin/retdec-bin2llvmir)
==5589==    by 0x478330: retdec::dwarfparser::DwarfFile::loadFileCUs() (in /home/peter/retdec-github/retdec/build/install/bin/retdec-bin2llvmir)
==5589==    by 0x4787D1: retdec::dwarfparser::DwarfFile::loadFile(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, retdec::fileformat::FileFormat*) (in /home/peter/retdec-github/retdec/build/install/bin/retdec-bin2llvmir)
==5589==    by 0x478A85: retdec::dwarfparser::DwarfFile::DwarfFile(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, retdec::fileformat::FileFormat*) (in /home/peter/retdec-github/retdec/build/install/bin/retdec-bin2llvmir)
[...]

PR #334, opened 2018-06-10
Fix of memory leak in Capstone2LlvmIrTranslator.

Alexander-Alekseev:
The 'insn' pointer was assigned values N-times without releasing the memory.
Fixes (part of?) memory leaks mentioned in #164, for example:

> ==22520== 11,046,552 (1,272,480 direct, 9,774,072 indirect) bytes in 5,302 blocks are definitely lost in loss record 1,245 of 1,245
> ==22520==    at 0x4C2CEDF: malloc (vg_replace_malloc.c:299)
> ==22520==    by 0x693D53: cs_malloc (in retdec-bin2llvmir)
> ==22520==    by 0x64EA29: retdec::capstone2llvmir::Capstone2LlvmIrTranslator::translate(std::vector<unsigned char, std::allocator<unsigned char> > const&, retdec::utils::Address, llvm::IRBuilder<llvm::ConstantFolder, llvm::IRBuilderDefaultInserter>&, bool) (capstone2llvmir.cpp:881)
> ==22520==    by 0x2C386B: retdec::bin2llvmir::Decoder::doDecoding() (decoder.cpp:1324)
> ==22520==    by 0x2BE092: retdec::bin2llvmir::Decoder::run() (decoder.cpp:109)
> ==22520==    by 0x2BDFDC: retdec::bin2llvmir::Decoder::runCatcher() (decoder.cpp:69)
> ==22520==    by 0x2BDF66: retdec::bin2llvmir::Decoder::runOnModule(llvm::Module&) (decoder.cpp:49)
> ==22520==    by 0x10BDB60: llvm::legacy::PassManagerImpl::run(llvm::Module&) (in retdec-bin2llvmir)
> ==22520==    by 0x201B24: _main(int, char**) (bin2llvmir.cpp:465)
> ==22520==    by 0x201E95: main (bin2llvmir.cpp:483)

### DIFF
--- a/src/capstone2llvmir/capstone2llvmir_impl.cpp
+++ b/src/capstone2llvmir/capstone2llvmir_impl.cpp
@@ -170,6 +170,7 @@ Capstone2LlvmIrTranslator_impl<CInsn, CInsnOp>::translate(
 			return res;
 		}
 
+		cs_free(insn, 1); // release previously allocated memory
 		insn = cs_malloc(_handle);
 
 		// TODO: hack, solve better.


PeterMatula:
`insn` pointer is stored to `TranslationResult::insns` so that the capstone2llvmir users can access both LLVM IR instructions and associated Capstone instructions. It is up to the users to free these Capstone pointers once they do not need them. If it gets freed like you propose, users would segfault.

The reason why decompiler did not segfault is that `Capstone2LlvmIrTranslator_impl::translate()` is not even used there, `Capstone2LlvmIrTranslator_impl::translateOne()` is used instead - and on successful translation, there is no `cs_free()`.

Maybe this could be solved better, like taking an extra argument that would cause `insn` to be freed if user does not need it. Or implementing another API function, etc. Bin2llvmir however needs access to Capstone instructions, so they cannot be freed at the time of translation. If there are leaks, more complex solution is needed.

PR #336, opened 2018-06-10
Bin2llvmir improvements

Alexander-Alekseev:
Several fixes for the bin2llvmir project.


Alexander-Alekseev:
Offtopic: please let me know if number of commits in a row isn't a good way. I will 'squash' them.

s3rvac:
> Offtopic: please let me know if several commits in a row isn't a good way. I will 'squash' them.

This is totally fine. It is better to create atomic commits with good commit messages than huge commits with unrelated changes and generic commit messages that tell you nothing about the reason why the change was necessary.

PeterMatula:
Thanks, i will look into it, but not right away.

PeterMatula:
Looks good.

There were some good catches in there. May I ask how did you find them? Did you just go over the code yourself, or compiler warnings, or used some code analysis tool?

Alexander-Alekseev:
For changes in these commits I used PVS-Studio (https://www.viva64.com/en/pvs-studio/).

PR #321, opened 2018-06-03
Implemented computation of export hashes

JakubPruzinec:
#121

- impHashBytes relict removed
- expHashes added in both, fileformat and fileinfo
- plain/json getters for presentation modified accordingly

expHashes are calculated after the table gets sorted alphabetically. ordinals are not converted to names rather to form of "ord<ord_num>"

PR #344, opened 2018-06-24
Resource icon hash computation

JakubPruzinec:
fileformat support for resource icon hashes calculation. the logic is implemented in computeIconHashes. fileinfo wrappers modified accordingly. fileinfo representation of resource table has been replaced. minor optimalization modifications were added.


metthal:
Why does your PR contain changes from #321? Please keep them separate.

PR #345, opened 2018-06-26
had to run those commands to be able to start building

jazzl0ver:


PeterMatula:
This has to do with the configuration of your system in order to use cmake, VS compiler, etc. This is not the kind of thing we want in README. Moreover, those lines you added are terribly specific to your system, VS version, etc. There is no way that this would be the same on all the Windows systems.

jazzl0ver:
@PeterMatula , I understand. Can these be added to a wiki, for example? The point is that I (am not a Windows developer) have spent plenty of time to make it building on my system (which was a clean Windows EC2 instance). That might help somebody else..

PeterMatula:
What exactly was the problem? Which command out of these failed and why?
- `cmake .. -DCMAKE_INSTALL_PREFIX=<path> -G<generator>`
- `cmake --build . --config Release -- -m`
- `cmake --build . --config Release --target install`

How did those commands you added help?

jazzl0ver:
@PeterMatula , this is completely my mistake - I apologies. I blindly copied your version of the generator while having VS 2017 installed.

One thing that is still probably worth to be mentioned is a forward slash rule in the _path_. It's kinda non-intuitive for a windows user to specify paths with forward slash. Using intuitive backward slash leads at least to inability to install files in _path\share\retdec\support\ folder_, since _path_ becomes, for example, _c:retdec_, instead of _c:\retdec_, so the files get installed somewhere under msys64 tree.

PeterMatula:
I added a note to description of `CMAKE_INSTALL_PREFIX` in `README.md`.

I looked into it and it looks like this is dependent on command line you are using. When I run `cmake .. -DCMAKE_INSTALL_PREFIX=C:\tmp\tmp` and then print `CMAKE_INSTALL_PREFIX` in cmake I get:
* MSYS command line: `CMAKE_INSTALL_PREFIX = C:tmptmp`
* `cmd.exe`: `CMAKE_INSTALL_PREFIX = C:/tmp/tmp`

Using quotes will fix the problem `cmake .. -DCMAKE_INSTALL_PREFIX="C:\tmp\tmp"`:
* MSYS command line: `CMAKE_INSTALL_PREFIX = C:/tmp/tmp`
* `cmd.exe`: `CMAKE_INSTALL_PREFIX = C:/tmp/tmp`

jazzl0ver:
Awesome, thank you!

PR #348, opened 2018-07-03
Fix compiler warning (unused variable 'ai')

Maroc-OS:
There was a compiler warning because "ai" never used, So this is a quick fix for that.

Here is the example : 

```
[ 72%] Building CXX object src/fileinfo/CMakeFiles/retdec-fileinfo.dir/file_presentation/file_presentation.cpp.o
/Users/x/Development/retdec/src/bin2llvmir/optimizations/decoder/ir_modifications.cpp:478:16: warning: unused variable 'ai' [-Wunused-variable]
        else if (auto ai = AsmInstruction(_module, addr))

```


PeterMatula:
This got fixed in #348.

PR #349, opened 2018-07-03
Eliminate "lib*.a(filexx.cpp.o)" has no symbols on macOS

Maroc-OS:
On macOS there is some warnings about object that has no symbols in it, this is needed to be fixed to make building more clean and clear.

example of build log :
```
[ 94%] Linking CXX static library libretdec-bin2llvmir.a
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libretdec-bin2llvmir.a(idioms_borland.cpp.o) has no symbols
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,6 +34,14 @@ elseif(UNIX)
 	set(CMAKE_INSTALL_RPATH "$ORIGIN/../lib")
 endif()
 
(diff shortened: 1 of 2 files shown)


Maroc-OS:
After those, there will be no warnings at all on macOS compile time.

Maroc-OS:
@s3rvac could you take a look.

Issue #169, opened 2018-02-07
Some functions emitted as UNREACHABLE even when reachable/keep-unreachable-funcs is used


csnover:
I’ve attempted a decompilation of an older Win32 DLL which was compiled using Visual C++ 4 and detected appropriately by RetDec. 553 of 4015 detected functions are being decompiled as UNREACHABLE. For example, this function (simple C++ object constructor):

```
; function: function_100532ef at 0x100532ef -- 0x100533cb
0x100532ef:   55                     	push ebp
0x100532f0:   8b ec                  	mov ebp, esp
0x100532f2:   6a ff                  	push -1
0x100532f4:   68 b3 33 05 10         	push 0x100533b3
0x100532f9:   64 a1 00 00 00 00      	mov eax, dword ptr fs:[0]
0x100532ff:   50                     	push eax
0x10053300:   64 89 25 00 00 00 00   	mov dword ptr fs:[0], esp
0x10053307:   83 ec 04               	sub esp, 4
0x1005330a:   53                     	push ebx
0x1005330b:   56                     	push esi
0x1005330c:   57                     	push edi
0x1005330d:   89 4d f0               	mov dword ptr [ebp - 0x10], ecx
0x10053310:   8b 4d f0               	mov ecx, dword ptr [ebp - 0x10]
0x10053313:   e8 80 ef 01 00         	call 0x10072298
0x10053318:   c7 45 fc 00 00 00 00   	mov dword ptr [ebp - 4], 0
0x1005331f:   8b 45 f0               	mov eax, dword ptr [ebp - 0x10]
0x10053322:   c7 00 e8 16 08 10      	mov dword ptr [eax], 0x100816e8
0x10053328:   8b 45 f0               	mov eax, dword ptr [ebp - 0x10]
0x1005332b:   c7 40 30 00 00 00 00   	mov dword ptr [eax + 0x30], 0
0x10053332:   8b 45 f0               	mov eax, dword ptr [ebp - 0x10]
0x10053335:   c7 40 20 00 00 00 00   	mov dword ptr [eax + 0x20], 0
0x1005333c:   8b 45 f0               	mov eax, dword ptr [ebp - 0x10]
0x1005333f:   c7 40 34 00 00 00 00   	mov dword ptr [eax + 0x34], 0
0x10053346:   8b 45 f0               	mov eax, dword ptr [ebp - 0x10]
0x10053349:   c7 40 24 00 00 00 00   	mov dword ptr [eax + 0x24], 0
[...]

PR #391, opened 2018-09-11
Fix #41, #169 - Handle pointers with segment override correctly

### LINKED ISSUE #41: Completely fails to decompile functions using RAII
I have now tested both the 32bit and the 64bit precompiled versions on Windows 10 and it doesnt seem to work at all. I have tested simple programs compiled with Visual Studio 12 and 14 in debug and release ,with and without pdbs and with and without the -k switch. It seems to be able to decompile most of the library functions which have been linked into the program. It also decompiles some of the user supplied functions.
But every function using RAII for C++ objects decompiles to basically this:

`
abort();
// UNREACHABLE
`

Atleast it seems like RAII is the cause. Since getters and simple calculations not involving any objects are decompiled just fine. [Here](https://pastebin.com/VmKKpSgK) are some examples from 2 programs i decompiled with pdb information.

palant:
See my analysis of the issue in #41 (#169 seems to be a duplicate). capstone2llvmir currently ignores segment override on pointers like `fs:[0]`. This makes LLVM think that the code is handling null pointers and remove that code as unreachable.

First step was to treat segment overrides correctly, the matching LLVM concept is address spaces. Unfortunately, the only place resembling a documentation for the mapping between segments and address spaces is https://github.com/avast-tl/llvm/blob/725d0cee133c6ab9b95c493f05de3b08016f5c3c/lib/Target/X86/X86ISelDAGToDAG.cpp#L1427, that's what I implemented.

Having address spaces on pointers caused issues further down in the pipeline. bin2llvmir was always producing bitcasts for pointer casts, this had to be changed into address space casts when address spaces change.

And llvmir2hll didn't know address space casts. I made it treat them the same as bitcasts which should do in the short term. In fact, I don't really know how that `add byte ptr gs:[ebx], dh` instruction (the one that produced the address space cast) got into [this driver](https://github.com/avast-tl/retdec-regression-tests/tree/master/features/windrivers) and what it could be converted into.

### DIFF
--- a/src/llvmir2hll/llvm/llvmir2bir_converters/orig_llvmir2bir_converter/llvm_converter.cpp
+++ b/src/llvmir2hll/llvm/llvmir2bir_converters/orig_llvmir2bir_converter/llvm_converter.cpp
@@ -182,6 +182,7 @@ ShPtr<Expression> LLVMConverter::llvmConstantToExpression(llvm::Constant *c) {
 				return IntToPtrCastExpr::create(op, llvmTypeToType(ce->getType()));
 
 			case llvm::Instruction::BitCast:
+			case llvm::Instruction::AddrSpaceCast:
 				return BitCastExpr::create(op, llvmTypeToType(ce->getType()));
 
 			case llvm::Instruction::GetElementPtr:
--- a/src/bin2llvmir/utils/ir_modifier.cpp
+++ b/src/bin2llvmir/utils/ir_modifier.cpp
@@ -71,12 +71,23 @@ Value* convertToType(
 	{
 		if (constExpr)
 		{
-			conv = ConstantExpr::getBitCast(cval, type);
+			if (val->getType()->getPointerAddressSpace() == type->getPointerAddressSpace())
+				conv = ConstantExpr::getBitCast(cval, type);
+			else
+				conv = ConstantExpr::getAddrSpaceCast(cval, type);
 		}
 		else
 		{
-			auto* i = new BitCastInst(val, type, "");
-			conv = insertBeforeAfter(i, before, after);
+			if (val->getType()->getPointerAddressSpace() == type->getPointerAddressSpace())
+			{
+				auto* i = new BitCastInst(val, type, "");
+				conv = insertBeforeAfter(i, before, after);
+			}
+			else
+			{
+				auto* i = new AddrSpaceCastInst(val, type, "");
+				conv = insertBeforeAfter(i, before, after);
+			}
 		}
 	}
 	else if (val->getType()->isPointerTy() && type->isIntegerTy())
--- a/src/capstone2llvmir/x86/x86.cpp
+++ b/src/capstone2llvmir/x86/x86.cpp
@@ -988,9 +988,32 @@ llvm::Value* Capstone2LlvmIrTranslatorX86_impl::loadOp(
 		}
 		case X86_OP_MEM:
 		{
-// TODO: what to do with this?
-//			auto* segR = loadRegister(op.mem.segment, irb);
-//			assert(segR == nullptr);
+			// LLVM doesn't really document address spaces. However, judging by
+			// X86ISelDAGToDag.cpp, the mapping is 256 -> GS, 257 -> FS, 258 -> SS.
+			unsigned addressSpace;
+			switch (op.mem.segment)
+			{
+				case X86_REG_FS:
+				{
+					addressSpace = 256;
+					break;
+				}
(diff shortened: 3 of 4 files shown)


palant:
The regression test bugs.idioms-reg-couples.TestCryptoLocker failed with these changes. Any reason why this regression test verifies that the bug I am fixing here is still present?

src/llvmir2hll/llvm/llvmir2bir_converters/orig_llvmir2bir_converter/llvm_converter.cpp
```diff
@@ -182,6 +182,7 @@ ShPtr<Expression> LLVMConverter::llvmConstantToExpression(llvm::Constant *c) {
 				return IntToPtrCastExpr::create(op, llvmTypeToType(ce->getType()));
 
 			case llvm::Instruction::BitCast:
+			case llvm::Instruction::AddrSpaceCast:
```
s3rvac: Thank you for the PR. Changes in `capstone2llvmir` and `bin2llvmir` will be reviewed by @PeterMatula, but I have a comment concerning the change in `llvmir2hll`. I think we should add `case llvm::Instruction::AddrSpaceCast` also to the following two places:

1. `src/llvmir2hll/llvm/llvmir2bir_converters/orig_llvmir2bir_converter/llvm_converter.cpp` (in `LLVMConverter::visitCastInst()`):
    ```cpp
    case llvm::Instruction::BitCast:
        return BitCastExpr::create(op, llvmTypeToType(i.getType()));
    ```
2. `src/llvmir2hll/llvm/llvmir2bir_converters/new_llvmir2bir_converter/llvm_instruction_converter.cpp` (in `LLVMInstructionConverter::convertConstExprToExpression`):
    ```cpp
    case llvm::Instruction::BitCast:
        return convertCastInstToExpression<BitCastExpr>(*cExpr);
    ```
    Including adding a unit test into `tests/llvmir2hll/llvm/llvmir2bir_converters/new_llvmir2bir_converter/llvm_value_converter_tests/llvm_instruction_converter_tests.cpp`. This code is part of a new LLVMIR to BIR converter that is under development. To use it, you need to run `retdec-decompiler.py` with `--backend-llvmir2bir-converter new`.

PeterMatula:
Thank you for this. We knew that work with `nullptr` pointers cause problems in LLVM passes, but did not get around to analyze and solve it. Your analysis and solution will no doubt improve the output quality a lot.

I did not merge it directly to `master` yet, since while studying the issue I decided to make some more changes upon your code:
* Test your changes on a larger testsuite that we use internally.
* Make it clear in the output C, that load/store from a different address space is happening. Right now this source (compiled using MSVC):
```c
int main(void)
{
    DWORD x = __readfsdword (0x18);
//  mov         eax,dword ptr fs:[00000018h] 
}
```
gets decompiled to:
```c
int main(int argc, char ** argv) {
    return *(int32_t *)24;
}
```
Hex-Rays generates this:
```c
int __cdecl main(int argc, const char **argv, const char **envp)
{
  return __readfsdword(0x18u);
}
```
It looks like it generates the same functions MSVC uses ([__readfsbyte](https://msdn.microsoft.com/en-us/library/3887zk1s.aspx), [__writefsbyte](https://msdn.microsoft.com/en-us/library/xfh6tchw.aspx)).
We were thinking about 2 alternatives:
   * Generate these pseudo functions in `capstone2llvmir`, instead of the address space generation code you added.
   * Leave your code in `capstone2llvmir`, and only later transform it to pseudo function calls.

We decided for the second option, because it contains the load/store semantics for LLVM passes - there are actual `load` and `store` instructions that may be analyzed and modified. We can generate the pseudo functions at the end, only when no more optimizations are to be performed. We should make sure no `addrspace(xyz)` survives this transformation - all of them get replaced.
* Test it again on a larger testsuite that we use internally.

This will solve most/many problems described in #41 and #169, since it removes one source of work with `nullptr` pointers. However, it does not guarantee there are no such instructions, therefore the LLVM optimizations you found in your analysis might still be applied if something goes wrong in the code RetDec sends to LLVM passes. I'm thinking about some sanitizer pass, that would try to find potential problems and hotfix them before running potentially aggressive LLVM passes. What do you think?

Also, if you have some objections/ideas about the things I wrote, lets discuss them here.

palant:
No objections on my end. I don't know the product too well, so I took the easy route - feel free to implement a proper solution. And - sure, ensuring that LLVM won't simply remove code it deems unreachable would be great, but that's also beyond my capabilities.

PR #398, opened 2018-09-13
Fix #397 - bin2llvmir: Add more known terminating functions to the list

### LINKED ISSUE #397: _CxxThrowException not recognized as a terminating call
It seems that RetDec only has a static list of terminating functions:

https://github.com/avast-tl/retdec/blob/418b84ab8fc5976a5963f0dc9ca0791f50b02c99/src/bin2llvmir/providers/fileimage.cpp#L566-L573

I have a DLL importing `_CxxThrowException` from VCRUNTIME140.DLL. This function isn't on the list, so I have a bunch of functions where the end isn't recognized correctly and the following function isn't decoded (considered unreachable code, because of the `int 3` instructions between functions).

palant:
This adds `_CxxThrowException` to the list of terminating functions, which improves quality quite significantly for me (see #397).


palant:
I now added more terminating functions to that list, there were quite a few missing. Not making a difference in my case, but maybe in some scenarios it does.

PR #399, opened 2018-09-14
Fix #384 - bin2llvmir: Be more strict when merging values

### LINKED ISSUE #384: Decompilation incorrectly uses pointer instead of the value it points to
Running RetDec built from https://github.com/avast-tl/retdec/commit/d57764abb7e45de61a22e45df298161b280316e1. I'm looking at a function that starts like this in the `.dsm` file:

```
0x10049390:   55                                 	push ebp
0x10049391:   8b ec                              	mov ebp, esp
0x10049393:   83 ec 14                           	sub esp, 0x14
0x10049396:   53                                 	push ebx
0x10049397:   56                                 	push esi
0x10049398:   8b 71 04                           	mov esi, dword ptr [ecx + 4]
0x1004939b:   8b da                              	mov ebx, edx
0x1004939d:   57                                 	push edi
0x1004939e:   8b 39                              	mov edi, dword ptr [ecx]
0x100493a0:   3b fe                              	cmp edi, esi
0x100493a2:   0f 84 81 00 00 00                  	je 0x10049429 <function_10049390+0x99>
```

In the `.c` file I get the following:

```
int32_t function_10049390(int32_t a1, int32_t a2) {
    // 0x10049390
    int32_t v1; // bp-24
    int32_t v2 = &v1; // 0x10049393
    int32_t v3 = g4; // 0x10049396
    int32_t v4 = g8; // 0x10049397
    int32_t str = g5; // 0x10049398
    int32_t v5 = *(int32_t *)(str + 4); // 0x10049398
    g8 = v5;
    int32_t v6 = g7; // 0x1004939b
    int32_t v7 = v6; // ebx
    int32_t v8 = g6; // 0x1004939d
    g6 = str;
    int32_t result;
    int32_t v9; // 0x10049429
    int32_t v10; // 0x1004942a
    if (v5 == str) {
```

So `ecx` is turned into `g5`, `edi` into `str` and `esi` into `v5` - so far it all makes sense. However, unlike indicated in the C program, `edi` and `ecx` are *not* the same. Rather, `ecx` points to a structure with two pointers, string start and string end. This appears to be some optimization gone wrong. I would rather expect the following here:

```
    int32_t v5 = *(int32_t *)(g5 + 4); // 0x10049398
    int32_t str = *(int32_t *)g5; // 0x1004939e
```

palant:
This fixes the issues mentioned in #384. I merely removed the check that made load result and the load origin considered equivalent, this is definitely wrong. Maybe some other checks here are also wrong, I can't really tell.

### DIFF
--- a/src/bin2llvmir/analyses/symbolic_tree.cpp
+++ b/src/bin2llvmir/analyses/symbolic_tree.cpp
@@ -413,8 +413,7 @@ void SymbolicTree::_simplifyNode()
 		*this = std::move(ops[0].ops[0]);
 	}
 	else if (match(*this, m_Load(m_Value(val), &load))
-			&& (isa<AllocaInst>(llvm_utils::skipCasts(load->getPointerOperand()))
-			|| isa<GlobalVariable>(llvm_utils::skipCasts(load->getPointerOperand()))))
+			&& isa<AllocaInst>(llvm_utils::skipCasts(load->getPointerOperand())))
 	{
 		*this = std::move(ops[0]);
 	}


palant:
Massive test fallout :disappointed:
I guess that this change is a no-go. But maybe you could help me understand why it has such a massive impact.

PeterMatula:
I will look into this. I think there might be something there - the change is not entirely bad and might make sense. But it looks like many RetDec parts are relying on the original behavior - maybe they should not and it should be solved in a different manner.

s3rvac:
@PeterMatula Were you able to look into this?

PR #442, opened 2018-12-14
Add support for snaps

drastx:
Signed-off-by: Dragan Stancevic <ds@codeminutia.com>

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,0 +1,23 @@
+name: retdec
+version: "8e65c05"
+summary: RetDec
+description: |
+  RetDec is a retargetable machine-code decompiler based on LLVM.
+  The decompiler is not limited to any particular target architecture,
+  operating system, or executable file format.
+  Supported file formats are ELF, PE, Mach-O, COFF, AR (archive), Intel HEX, and raw machine code.
+  Supported architectures (32b only) are Intel x86, ARM, MIPS, PIC32, and PowerPC.
+
+confinement: strict
+
+apps:
+  retdec:
+    command: bin/retdec-decompiler.py
+
+parts:
+  retdec:
+    plugin: cmake
+    configflags:
+      - -DCMAKE_INSTALL_PREFIX=/
+    source-type: git
+    source: https://github.com/avast-tl/retdec.git


snapcraft.yaml
```diff
@@ -0,0 +1,23 @@
+name: retdec
+version: "8e65c05"
```
s3rvac: Shouldn't we use a stable version (e.g. `3.2`, which is the latest stable version of RetDec) instead of a particular commit in the current `master`?

snapcraft.yaml
```diff
@@ -0,0 +1,23 @@
...
+  retdec:
+    command: bin/retdec-decompiler.py
+
+parts:
+  retdec:
+    plugin: cmake
+    configflags:
+      - -DCMAKE_INSTALL_PREFIX=/
+    source-type: git
+    source: https://github.com/avast-tl/retdec.git
```
s3rvac: Does snap automatically checkout `version` (e.g. commit `8e65c05`) in the cloned repo? The source URL `https://github.com/avast-tl/retdec.git` points to the current `master`.

s3rvac: Were you able to build a snap package by using the submitted `snapcraft.yml` file? I was not yet able to try it, but what seems fishy to me is that RetDec has a lot of dependencies (see e.g. [Dockerfile](https://github.com/avast-tl/retdec/blob/master/Dockerfile)), and none of them are listed in the `snapcraft.yml` file.

snapcraft.yaml
drastx: I haven't tested it but it should be able to checkout a specific release tag i.e. 3.2

snapcraft.yaml
drastx: Yes, correct. I should be checking out that particular commit.

drastx: Yes, I have built the snap on my system. It's uploaded to the snap store  at:
https://snapcraft.io/retdec

I tested the snap build on my system, which did have all the build dependencies. But the point of a snap is that the dependencies get built in so it should work on other systems the same way.

snapcraft.yaml
s3rvac: Thank you for all the answers. Could I ask you to update the `snapcraft.yml` file so it builds the [v3.2](https://github.com/avast-tl/retdec/releases/tag/v3.2) version, which is the current stable release of RetDec?

drastx: Hey @s3rvac, sorry have been really busy and had no time do do this yet but I haven't forgotten.

s3rvac: That's alright ;-).

s3rvac:
Closing due to inactivity. If you find some time to implement the required changes (+ to update the `snapcraft.yaml` file to RetDec version 3.3, which is the current version), please re-open.

xloem:
There is a currently a snap for this in the snap store. It hasn't been updated since 2018.

PR #450, opened 2018-12-29
Remove a recursive function call with a iterative function

### LINKED ISSUE #383: LOG_TIME = ['/usr/bin/time', '-v'] in retdec-config.py option "-v" not valid on macOS
The version in /usr/bin/time on macOS does not accept the '-v' option. The workaround is to install gnu-time from homebrew which would then be under /usr/local/bin/gtime. If I get time, I may post a fix for this, but for now just reporting the issue.

Cyperghost:
- The X87 analysis function breaks by large executable files, with a stack overflow 

and
- Fix #383


scripts/retdec-config.py
```diff
@@ -87,5 +87,10 @@
 UNPACKER = os.path.join(INSTALL_BIN_DIR, 'retdec-unpacker')
 
 # Other.
-LOG_TIME = ['/usr/bin/time', '-v']
+
+if platform == "darwin":
+    # mac os x doesn't support the v flag
+    LOG_TIME = ['/usr/bin/time']
```
s3rvac: Although this fixes the `/usr/bin/time: invalid option -- 'v'` error on macOS, it does not represent a full solution to #383 as our scripts assume that `/usr/bin/time` is run with `-v`. Apart from other information, the `-v` parameter makes `/usr/bin/time` produce memory usage of the tracked program, which we then parse in our scripts. Without `-v`, the used memory will always be 0, which I believe is incorrect and thus this fix cannot be merged. What is your take on this @PeterMatula ?

Cyperghost: Thanks, I looked in the docs by Mac OS X and I found the parameter `-l`.
I will change it.

Cyperghost: But I found a package by `brew` the `gnu-time`, this will give the same output. 
Which would be better to integrate?

s3rvac:
@PeterMatula Can you please take a look at this PR? It has been opened for more than a month now.

PR #484, opened 2019-02-03
Added option -l --list to retdec-bin2pat.

### LINKED ISSUE #472: retdec-bin2pat fails with "FileNotFoundError: The filename or extension is too long"
I am attempting to create a yara signature from the open-source ps2sdk through the retdec-signature-from-library-creator.py script. Upon running it I receive the error

FileNotFoundError: [WinError 206] The filename or extension is too long

Upon modifying retdec-utils.py to print out the arguments being passed to retdec-bin2pat.exe I can see that it is indeed too long. After copying the command into notepad++ I can see that it is a total of 118,999 characters long.

I did look into a way to get around this. You can see [here](https://support.microsoft.com/en-us/help/830473/command-prompt-cmd-exe-command-line-string-limitation) that Microsofts recommends modifying programs that require long command lines to use a file that contains the parameter information instead. A good solution would appear to give an option to retdec-bin2pat to do so and have retdec-utils.py automatically pass them this way if they exceed a certain length.

astrelsky:
The added option -l --list is being used to pass the list of objects
from retdec-signature-from-library-creator.py as a json file.
Resolves #472 

I am uncertain if the change in retdec-bin2pat's help output correctly
follows the Utility Conventions. If it does not, providing a proper
reference would be appreciated.

Regards,
Andrew Strelsky

### DIFF
--- a/scripts/retdec-signature-from-library-creator.py
+++ b/scripts/retdec-signature-from-library-creator.py
@@ -82,6 +82,7 @@ class SigFromLib:
 
         dir_name = os.path.dirname(os.path.abspath(self.args.output))
         self.tmp_dir_path = tempfile.mkdtemp(dir=dir_name)
+        self.object_list_path = os.path.join(self.tmp_dir_path, 'object-list.txt')
 
         if self.args.ignore_nops:
             self.ignore_nop = '--ignore-nops'
@@ -126,7 +127,10 @@ class SigFromLib:
             # Extract patterns from library.
             pattern_file = os.path.join(self.tmp_dir_path, lib_name) + '.pat'
             pattern_files.append(pattern_file)
-            _, result, _ = cmd.run_cmd([config.BIN2PAT, '-o', pattern_file] + objects, discard_stdout=True, discard_stderr=True)
+            with open(self.object_list_path, 'w') as object_list:
+                for item in objects:
+                    object_list.write(item+'\n')
+            _, result, _ = cmd.run_cmd([config.BIN2PAT, '-o', pattern_file, '-l', self.object_list_path], discard_stdout=True, discard_stderr=True)
 
             if result != 0:
                 self.print_error_and_cleanup('utility bin2pat failed when processing %s' % lib_path)
--- a/src/bin2pat/bin2pat.cpp
+++ b/src/bin2pat/bin2pat.cpp
@@ -28,13 +28,16 @@ void printUsage(
 	std::ostream &outputStream)
 {
 	outputStream << "Usage: bin2pat [-o OUTPUT_FILE] [-n NOTE]"
-		<< " INPUT_FILE [INPUT_FILE...]\n\n"
+		<< " <INPUT_FILE [INPUT_FILE...] | -l LIST_FILE>\n\n"
 		<< "-o --output OUTPUT_FILE\n"
 		<< "    Output file path (if not given, stdout is used).\n"
 		<< "    If multiple paths are given, only last one is used.\n\n"
 		<< "-n --note NOTE\n"
 		<< "    Optional note that will be added to all rules.\n"
-		<< "    If multiple notes are given, only last one is used.\n\n";
+		<< "    If multiple notes are given, only last one is used.\n\n"
+		<< "-l --list LIST_FILE\n"
+		<< "    Optionally pass the list of input files as a text file.\n"
+		<< "    This is useful for a large number of input files.\n\n";
 }
 
 void printErrorAndDie(
@@ -79,6 +82,35 @@ void processArgs(
 			else {
 				needValue(args[i]);
 				return;
+			}
+		}
+		else if (args[i] == "-l" || args[i] == "--list") {
+			// Ensure -l --list is not the last thing in args
+			if (&args[i] == &args.back()) {
+				printErrorAndDie("input file missing");
+				return;
+			}
+
+			std::ifstream inputObjects(args[++i]);
+			std::string object;
+
+			if (!inputObjects) {
+				printErrorAndDie("LIST_FILE '" + args[i]
+					+ "' is not a valid file");
+				return;
+			}
+			// Read LIST_FILE until EOF
+			while (std::getline(inputObjects, object)) {
+				// Ensure file exists before proceeding
+				if(!FilesystemPath(object).isFile()) {
+					printErrorAndDie("argument '" + args[i]
+						+ "' contains the filename '" + object
+						+ "' which is not a valid file");
+					return;
+				}
+				else {
+					inPaths.push_back(object);
+				}
 			}
 		}
 		else {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 # dev
 
+* Fix: Added option to `retdec-bin2pat` to have the objects list passed through a text file.
+Modified `retdec-signature-from-library-creator.py` to pass objects list as a text file. ([#472](https://github.com/avast-tl/retdec/issues/472)).
 * New Feature: Added presentation of imported types and TypeRef hashes for .NET binaries ([#363](https://github.com/avast-tl/retdec/issues/363), [#364](https://github.com/avast-tl/retdec/issues/364), [#428](https://github.com/avast-tl/retdec/issues/428)).
 * New Feature: Added computation and presentation of icon hashes for exact and also similarity matching in PE files ([#339](https://github.com/avast-tl/retdec/issues/339)).
 * Enhancement: Added support for build and run on FreeBSD and potentially on other BSD OSes ([#476](https://github.com/avast-tl/retdec/pull/476)).


src/bin2pat/bin2pat.cpp
```diff
@@ -81,6 +85,30 @@ void processArgs(
...
+				{
+					for (int j = 0; j < jsonIn.size(); j++)
+						inPaths.push_back(jsonIn[j].asCString());
+					jsonFile.close();
+				}
+				else
+				{
+					jsonFile.close();
+					printErrorAndDie("JSON_FILE '" + args[i]
+					+ "' could not be parsed");
```
s3rvac: Nitpick: Missing indentation. It should be
```cpp
printErrorAndDie("JSON_FILE '" + args[i]
	+ "' could not be parsed");
```

src/bin2pat/bin2pat.cpp
```diff
@@ -81,6 +85,30 @@ void processArgs(
 				return;
 			}
 		}
+		else if (args[i] == "-j" || args[i] == "--json") {
+			Json::Value jsonIn;
+			Json::Reader reader;
+			std::ifstream jsonFile(args[++i], std::ifstream::binary);
```
s3rvac: What if `-j` or `--json` is the very last argument? Wouldn't this crash when trying to access `++i`? Shouldn't we also check that there is one more argument after `-j`/`--json`?

astrelsky:
@s3rvac Thank you for taking the time to put so much detail into reviewing this. I must be honest, when I awoke to the email notification I had an enormous grin on my face. I have not seen such a list of required changes since I took my introductory to C++ course. I will amend the commit this afternoon.

With regard to using json, I must confess that this is a case of using it because it was there. I should have followed the golden rule of 'keep it simple, stupid'. Thinking about it again, its use just adds overhead and consumes more memory than necessary. Considering that a user is unlikely to try to pass a huge list of objects manually, is it necessary to even have the option appear when the help is displayed? Do you have a suggestion for the command-line option?

astrelsky:
I have amended the commit. I have also pushed another commit to resolve issue #490 . I noticed when checking the commit changes to resolve #472 that the isFile check in retdec-bin2pat did not actually stop the filename from being added to the inPaths list.

If having the additional commit is an issue I will drop it and update.

src/bin2pat/bin2pat.cpp
```diff
@@ -28,13 +29,16 @@ void printUsage(
 	std::ostream &outputStream)
 {
 	outputStream << "Usage: bin2pat [-o OUTPUT_FILE] [-n NOTE]"
-		<< " INPUT_FILE [INPUT_FILE...]\n\n"
+		<< " <INPUT_FILE [INPUT_FILE...] | JSON_FILE>\n\n"
```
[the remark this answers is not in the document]
astrelsky: I missed this one in the revised changes. Once the checks complete it will again be revised.

src/bin2pat/bin2pat.cpp
```diff
@@ -81,6 +85,30 @@ void processArgs(
...
+				printErrorAndDie("JSON_FILE '" + args[i]
+					+ "' does not exist");
+				return;
+			}
+			else {
+				if (reader.parse(jsonFile, jsonIn, false))
+				{
+					for (int j = 0; j < jsonIn.size(); j++)
+						inPaths.push_back(jsonIn[j].asCString());
+					jsonFile.close();
```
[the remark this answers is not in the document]
astrelsky: It turns out that it is indeed closed in the destructor.

scripts/retdec-signature-from-library-creator.py
```diff
@@ -82,6 +84,7 @@ def _check_arguments(self):
 
         dir_name = os.path.dirname(os.path.abspath(self.args.output))
         self.tmp_dir_path = tempfile.mkdtemp(dir=dir_name)
+        self.tmp_fd, self.tmp_file_path = tempfile.mkstemp(dir=dir_name, text=True)
```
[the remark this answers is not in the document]
astrelsky: I have instead setup a text file named "objectLists.txt" that resides within the created temporary directory. Is this more acceptable?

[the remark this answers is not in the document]
s3rvac: Yes. I would just rename the file to `object-list.txt` as we generally use either dashes (`-`) or underscores (`_`) in file names.

src/bin2pat/bin2pat.cpp
```diff
@@ -81,6 +83,37 @@ void processArgs(
 				return;
 			}
 		}
+		else if (args[i] == "-l" || args[i] == "--list") {
+			// Ensure -l --list is not the last thing in args
+			if (&args[i] == &args.back()) {
```
s3rvac: Here, I think the code would be more readable if we just compared `i` with the size of `args`:
```cpp
if (i == args.size() - 1) {
```
The comparison of addresses feels very unnatural and kept me thinking really hard whether the check is actually correct (I am still not entirely convinced that it is correct). What do you think?

src/bin2pat/bin2pat.cpp
```diff
@@ -81,6 +83,37 @@ void processArgs(
...
+				printErrorAndDie("input file missing");
+				return;
+			}
+
+			std::ifstream inputObjects(args[++i]);
+			std::string object;
+
+			if (inputObjects.is_open()) {
```
s3rvac: I suggest moving the `if` and `else` clauses around to make the code more readable and reduce the indentation level. We could do this by converting
```cpp
if (inputObjects.is_open()) {
    // Read LIST_FILE until EOF
    while (inputObjects >> object) {
        // Ensure file exists before proceeding
        if(!FilesystemPath(object).isFile()) {
            printErrorAndDie("argument '" + args[i]
                + "' contains the filename '" + object
                + "' which is not a valid file");
            return;
        }
        else {
            inPaths.push_back(object);
        }
    }
}
else {
    printErrorAndDie("LIST_FILE '" + args[i]
        + "' is not a valid file");
    return;
}
```
into
```cpp
if (!inputObjects) {
    printErrorAndDie("LIST_FILE '" + args[i]
        + "' cannot be opened");
    return;
}

// Read LIST_FILE until EOF
while (inputObjects >> object) {
    // Ensure file exists before proceeding
    if(!FilesystemPath(object).isFile()) {
        printErrorAndDie("argument '" + args[i]
            + "' contains the filename '" + object
            + "' which is not a valid file");
        return;
    }
    else {
        inPaths.push_back(object);
    }
}
```
What do you think about this? This is just a suggestion, so if you like the original version, feel free to keep it.

src/bin2pat/bin2pat.cpp
```diff
@@ -81,6 +83,37 @@ void processArgs(
...
+				// Read LIST_FILE until EOF
+				while (inputObjects >> object) {
```
s3rvac: [`std::istream::operator>>()`](https://en.cppreference.com/w/cpp/io/basic_istream/operator_gtgt) stops reading after encountering the first whitespace character. Therefore, if a file path contains spaces, it will be split into multiple paths. This is rare, but might happen as some users/applications put spaces into file names. Actually, [one of our tests failed](https://retdec-tc.avast.com/viewLog.html?buildId=1183&buildTypeId=Retdec_LinuxBuild&tab=buildLog&_focus=3450) exactly because of this issue.

To illustrate, an input file containing a single line
```
my file with spaces.txt
```
would be parsed into the following objects:
```
my
file
with
spaces.txt
```

Therefore, to be safe, we should switch to [`std::getline()`](https://en.cppreference.com/w/cpp/string/basic_string/getline) to read the objects from the input file. This can be simply done by replacing
```cpp
while (inputObjects >> object) {
```
with
```cpp
while (std::getline(inputObjects, object)) {
```

src/bin2pat/bin2pat.cpp
```diff
@@ -9,7 +9,6 @@
 #include <iostream>
 #include <ostream>
 #include <vector>
-
```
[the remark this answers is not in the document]
astrelsky: That's interesting, I will see to it that it is removed.

src/bin2pat/bin2pat.cpp
astrelsky: The reason I used a comparison of addresses here is because it is absolute. I don't see how this could potentially fail and while I didn't go to the ends of the Earth to break it, it did work.

Using a comparison of addresses ensures thst, at the time the program counter reaches that section of code, that the argument being read is currently the last argument in args.

I can change it for readability reasons and the difference breaks down to only one instruction due to the subtraction.

The only reason I did not use your above suggestion is because it is comparing the current index count to the size of args in relation to it's index as opposed to comparing if the current argument being evaluated is the final argument in args.

src/bin2pat/bin2pat.cpp
astrelsky: This, or something similar to it, will be implemented for readability when fixing the issue regarding inputObjects >> object.

src/bin2pat/bin2pat.cpp
astrelsky: I was mistakenly under the impression that the delimiter for reading a string from a stream was a new line character. I will resolve this with the above suggestion.

PR #491, opened 2019-02-05
Fixed FilesystemPath::isFile return value.

### LINKED ISSUE #490: FilesystemPath::isFile returning true on non existent file
I noticed that each implementation of FilesystemPath::isFile returns !isDirectory. The isDirectory functions appear to check whether the path is valid or exists and returns false if it is not. This causes instances of isFile to return true when the path may not even exist.

In the docstring it states "The path doesn't have to be loaded." Does this imply that it may not exist? If this is the case there may be instances where a non-existent file is not caught.

astrelsky:
Resolves #490 

@metthal I went ahead and implemented your suggestions. I also split it into a separate branch and removed these changes from the other pull request as I should have done in the first place.

Regards,
Andrew Strelsky

### DIFF
--- a/src/utils/filesystem_path.cpp
+++ b/src/utils/filesystem_path.cpp
@@ -114,7 +114,7 @@ public:
 
 		subpaths.clear();
 		HANDLE hFnd = FindFirstFile(examineDir.c_str(), &ffd);
-		if (hFnd == reinterpret_cast<HANDLE>(-1))
+		if (hFnd == INVALID_HANDLE_VALUE)
 			return false;
 
 		do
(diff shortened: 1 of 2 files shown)


metthal:
Thank you! LGTM. I'll just wait for TeamCity to finish all builds and then I will merge it. AppVeyor will probably fail as we have problems with timeouts in it right now but no worries.

s3rvac:
The [AppVeyor](https://ci.appveyor.com/project/avast-tl/retdec/builds/22153959) build failure can be ignored as it timeouted (again...). We will have to do something with it.

src/utils/filesystem_path.cpp
```diff
@@ -141,7 +141,11 @@ class FilesystemPathImplWindows : public FilesystemPathImpl
 
 	virtual bool isFile() override
 	{
-		return !isDirectory();
+		WIN32_FIND_DATA ffd;
+		if (FindFirstFile(_path.c_str(), &ffd) == reinterpret_cast<HANDLE>(-1))
```
s3rvac: Couldn't we replace `== reinterpret_cast<HANDLE>(-1)` with the more readable `== INVALID_HANDLE_VALUE`? I have taken it from [this example](https://docs.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-findfirstfilea#examples).

src/utils/filesystem_path.cpp
```diff
@@ -141,7 +141,11 @@ class FilesystemPathImplWindows : public FilesystemPathImpl
 
 	virtual bool isFile() override
 	{
-		return !isDirectory();
+		WIN32_FIND_DATA ffd;
```
s3rvac: Instead of `FindFirstFile()`, wouldn't it be easier and more straightforward to use just `GetFileAttributes()`?
```cpp
virtual bool isFile() override
{
	auto attrs = GetFileAttributes(_path.c_str());
	return (attrs != INVALID_FILE_ATTRIBUTES &&
		(attrs & FILE_ATTRIBUTE_DIRECTORY) == 0);
}
```
I am not a Windows user, but `GetFileAttributes()` kind of more resembles the POSIX `stat()` function.

src/utils/filesystem_path.cpp
astrelsky: Yes. I have gone ahead and applied this to isDirectory as well.

src/utils/filesystem_path.cpp
metthal: I agree with @s3rvac that `GetFileAttributes` makes more sense here than `FindFirstFile`.

astrelsky: I have implemented it as suggested and did a forced update. At first I thought that the above would incorrectly return true on INVALID_FILE_ATTRIBUTES since it is defined as -1. However, I see that the logical && and == causes it to indeed function properly.

metthal:
Thank you for your contribution. I've merged your changes.

PR #492, opened 2019-02-06
Remove more than half the stack usage of CFG traversal

dberlin:
A lot of recursive stack space is used walking the statements in order, which is not necessary.

### DIFF
--- a/src/llvmir2hll/graphs/cfg/cfg_traversal.cpp
+++ b/src/llvmir2hll/graphs/cfg/cfg_traversal.cpp
@@ -158,7 +158,7 @@ bool CFGTraversal::getCurrRetVal() const {
 }
 
 /**
-* @brief A recursive implementation of performTraversal().
+* @brief A non-recursive implementation of performTraversal().
 *
 * @param[in] node Node to be traversed.
 * @param[in] stmtIter Iterator to a statement in @a node to be checked.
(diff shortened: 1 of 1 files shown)


dberlin:
Sorry, missed staging a hunk on the reverse one, one minute

src/llvmir2hll/graphs/cfg/cfg_traversal.cpp
```diff
@@ -168,7 +168,7 @@ bool CFGTraversal::getCurrRetVal() const {
 */
```
s3rvac: The descriptions of the two modified functions are
```
* @brief A recursive implementation of performTraversal().
```
```
* @brief A recursive implementation of performReverseTraversal().
```
I guess we should remove the word _recursive_ from them as they are no longer recursive.

dberlin:
Sorry, i'll fix.
I have a follow on patch that removes all the recursion (the version here still takes about 30+ gig of stack on some of my test cases), and i updated the comments in there :)

dberlin:
I updated the comment. Note that they are still recursive, as traverseNodeSuccessors will recurse back into calling us.

The traversal of CFG nodes as implemented is quite weird in actual practice. I rewrote it in a followup, i'll include more details there.

src/llvmir2hll/graphs/cfg/cfg_traversal.cpp
dberlin: Changed to non-recursive

s3rvac:
That AppVeyor build failure is due to a timeout and has nothing to do with this PR. We will have to figure what to do with it.

Issue #479, opened 2019-01-30
Regression tests timeout from time to time


s3rvac:
Some of our regression tests timeout from time to time. For an example, see [this macOS build](https://retdec-tc.avast.com/viewLog.html?buildId=1083&buildTypeId=Retdec_MacBuild&tab=buildLog). I have seen the timeouts on both macOS and Linux, but maybe this is also an issue on Windows. I strongly believe that these timeouts are caused by non-determinism in our tools (#209, #129). Instead of keeping disabling failing tests (#266), we should try to debug this.

[Here](https://github.com/avast-tl/retdec/files/2811071/TestNewARM32Signatures-outputs.zip) is an example of two `.ll` files produced by `bin2llvmir` in two separate runs of the `features.new-static-code-signatures.TestNewARM32Signatures` regression test on my Linux machine. As you can see, they differ on quite a lot of places. @PeterMatula, do you have an idea what is causing these differences?

s3rvac:
More timeouts on macOS: https://retdec-tc.avast.com/viewLog.html?buildId=1249&buildTypeId=Retdec_MacBuild&tab=buildLog&_focus=5200

@PeterMatula Can you please take a look at this issue?

PeterMatula:
Fixed in 55d46b5, all previously disabled tests enabled in https://github.com/avast-tl/retdec-regression-tests/commit/87dd055a96d4a512204082ec1dabedaa2505b7b4.

Issue #209, opened 2018-02-08
Hunting for nondeterminism in our tools


PeterMatula:
From time to time, we notice nondeterminism in some of our tools. We try to fix it once it is identified. It would be nice to be able to automatically and periodically test for it. We should ideally test:
- Diffs in several runs of the same build.
- Diffs in runs of different builds (compilers, options, etc.).
- Diffs in runs on different systems (Linux/Windows).
Moreover, there might be some inherent nondeterminisms - time stamps, absolute paths, etc. 

Right now, I have no idea how to do something like this. If anyone have any suggestion, please comment.

PeterMatula:
Commits disabling regression tests on macOS due to non-determinism:
* https://github.com/avast-tl/retdec-regression-tests/commit/93f9d667a0f9c01d0ab1dd5988964d460f80a5bd
* https://github.com/avast-tl/retdec-regression-tests/commit/46f3b6684aabbc14a9e128adf5ede1d08e5e66da

PR #495, opened 2019-02-07
Rewrite recursive cfg traversal to non-recursive

dberlin:
This removes the rest of the stack usage in the *forward* CFG traversal.

On very large BB testcases i have (millions of BBs), this reduces stack usage to essentially nothing.

This implementation is also somewhat faster for other reasons:
The recursive implementation will touch a lot more nodes due to how it recurses - it will touch each node, start processing the successors, and then recurse on itself (which then processes successors). Depending on CFG structure, when it pops back up, it will now touch a whole bunch of successors that it already processed (it then later discovers it already checked them).
This is fairly bad cache behavior, etc.

This implementation will not, they are cut off before any other processing is done.
It is a minor thing for most CFGs, of course.

I have not updated the reverse traversal.  In practice, the next step really should be to turn this into a generic depth first iterator rather than duplicate the code.

The code is based on code i wrote for LLVM's depth first iterator (http://llvm.org/doxygen/DepthFirstIterator_8h_source.html), and meant to be able to be turned into one fairly easily.

I just don't have time at the moment to do it :)

### DIFF
--- a/src/llvmir2hll/graphs/cfg/cfg_traversal.cpp
+++ b/src/llvmir2hll/graphs/cfg/cfg_traversal.cpp
@@ -4,6 +4,10 @@
 * @copyright (c) 2017 Avast Software, licensed under the MIT license
 */
 
+#include <stack>
+#include <tuple>
+#include <unordered_set>
+
 #include "retdec/llvmir2hll/graphs/cfg/cfg_traversal.h"
 #include "retdec/llvmir2hll/ir/empty_stmt.h"
 #include "retdec/llvmir2hll/support/debug.h"
@@ -13,6 +17,97 @@ using retdec::utils::hasItem;
 
 namespace retdec {
 namespace llvmir2hll {
+
+namespace {
+
+template <class Graph>
+class df_iterator : public std::iterator<std::forward_iterator_tag,
+					 ShPtr<typename Graph::Node>>
+{
+	using Node = ShPtr<typename Graph::Node>;
+	using super = std::iterator<std::forward_iterator_tag, Node>;
+	// The stack nodes consist of the current CFG node, a statement iterator
+	// storing the current place in the node's statements, and an iterator
+	// storing the current place in the node's successor list.  Statement
+	// place is only needed because we may be called with a statement that
+	// doesn't point to the beginning.
+	using StackNodeType = std::tuple<Node, typename Graph::stmt_iterator, typename Graph::succ_iterator>;
+	using ReturnNodeType = std::pair<Node, typename Graph::stmt_iterator>;
+	std::stack<StackNodeType> visitStack;
+	std::unordered_set<Node> visitedNodes;
+private:
+	inline df_iterator(Node node, typename Graph::stmt_iterator stmtIter) {
+		visitedNodes.insert(node);
+		visitStack.emplace(node, stmtIter, node->succ_begin());
+	}
+	inline df_iterator(Node node) : df_iterator(node, node->stmt_begin()) {
+	}
+	inline df_iterator() = default; // End is when stack is empty
+	inline void moveToNextNode() {
+		// Note that we directly mutate the successor iterators that are
+		// on the stack, because we can't pop it until we are done.
+		do {
+			auto node = std::get<0>(visitStack.top());
+			auto &succPlace = std::get<2>(visitStack.top());
+			// Now push the next successor on the stack to make it
+			// get visited.
+			while (succPlace != node->succ_end()) {
+				auto succNode((*succPlace)->getDst());
+				// This mutates the actual in-stack succ
+				// iterator
+				++succPlace;
+				// Push this node on the stack to be visited if
+				// it hasn't been visited.
+				if (visitedNodes.insert(succNode).second) {
+					visitStack.emplace(succNode, succNode->stmt_begin(),
+							   succNode->succ_begin());
+					return;
+				}
+			}
+			// Otherwise, we need to pop up a level because we are done
+			// completely with the stack node.
+			visitStack.pop();
+		} while (!visitStack.empty());
+	}
+public:
+	using pointer = typename super::pointer;
+
+	// Provide static begin and end methods as our public "constructors"
+	static df_iterator<Graph> begin(const Node &N) { return df_iterator(N); }
+	static df_iterator<Graph> begin(const Node &N,
+					typename Graph::stmt_iterator stmtIter) {
+		return df_iterator(N, stmtIter);
+	}
+
+	static df_iterator<Graph> end(const Node &N) { return df_iterator(); }
+
+	bool operator==(const df_iterator &x) const {
+		return visitStack == x.visitStack;
+	}
+	bool operator!=(const df_iterator &x) const { return !(*this == x); }
+
+	ReturnNodeType operator*() const { auto &topNode = visitStack.top();
+		return std::make_pair(std::get<0>(topNode), std::get<1>(topNode)); }
+
+	df_iterator &operator++() { // Preincrement
+		moveToNextNode();
+		return *this;
+	}
+
+  // Skips all children of the current node.  Note that this may cause the
+  // iterator to be at end when it is done.
+  df_iterator &skipChildren() {
+    visitStack.pop();
+    return *this;
+  }
+
+  df_iterator operator++(int) { // Postincrement
+    df_iterator tmp = *this;
+    ++*this;
+    return tmp;
+  }
+};
+} // anonymous namespace
 
 /**
 * @brief Constructs a new traverser.
@@ -75,18 +170,28 @@ bool CFGTraversal::performTraversalFromSuccessors(ShPtr<Statement> stmt) {
 
 	// Initialization.
 	stopTraversal = false;
-
 	CFG::StmtInNode stmtInNode(cfg->getNodeForStmt(stmt));
 	ASSERT_MSG(stmtInNode.first,
 		"the statement `" << stmt << "` is not in the CFG");
 
+	auto retVal = getEndRetVal();
 	if (stmtInNode.second != stmtInNode.first->stmt_end()) {
 		// It has a successor in the same node, so start traversing from the
 		// successor.
 		return performTraversalImpl(stmtInNode.first, ++stmtInNode.second);
+	} else {
+		auto node = stmtInNode.first;
+		// It is the last statement in the node, so traverse all node successors.
+		// For each outgoing edge...
+		for (auto i = node->succ_begin(), e = node->succ_end(); i != e; ++i) {
+			ShPtr<CFG::Node> dstNode((*i)->getDst());
+			retVal = combineRetVals(retVal, performTraversalImpl(node, node->stmt_end()));
+			if (stopTraversal) {
+				break;
+			}
+		}
 	}
-	// It is the last statement in the node, so traverse all node successors.
-	return traverseNodeSuccessors(stmtInNode.first);
+	return retVal;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


s3rvac:
Thank you for the PR! :+1:  I will review it and will let you know.

As for the failed continuous integrations:
* The TeamCity checks have failed on a timeout in a regression test. This is a generic issue and not related to this PR, see #479 (CC @PeterMatula).
* The Linux TravisCI build [has failed](https://travis-ci.org/avast-tl/retdec/jobs/490297897#L2276) to build `src/llvmir2hll/graphs/cfg/cfg_traversal.cpp`:
    ```
    src/llvmir2hll/graphs/cfg/cfg_traversal.cpp: In member function ‘bool retdec::llvmir2hll::CFGTraversal::performTraversalImpl(retdec::llvmir2hll::ShPtr<retdec::llvmir2hll::CFG::Node>, retdec::llvmir2hll::CFG::stmt_iterator)’:
    src/llvmir2hll/graphs/cfg/cfg_traversal.cpp:191:54: error: converting to ‘std::stack<...>::value_type’ from initializer list would use explicit constructor ‘constexpr std::tuple<>::tuple(...) [...]’
    ```
    Could you please verify? It uses GCC 4.9, which is the oldest GCC version that we support. If you are unable to reproduce this, let me know and I will take a look.

src/llvmir2hll/graphs/cfg/cfg_traversal.cpp
```diff
@@ -4,6 +4,8 @@
 * @copyright (c) 2017 Avast Software, licensed under the MIT license
```
s3rvac: Several tests in our regression test suite in TeamCity have failed (e.g. [these tests](https://retdec-tc.avast.com/viewLog.html?buildId=1280&buildTypeId=Retdec_LinuxBuild&tab=buildLog&_focus=3734)). This seems to be caused by your changes, which indicates that there is probably a bug somewhere. I will try to investigate.

BTW, you can ignore the timeout in the `features.new-static-code-signatures.TestNewARM32Signatures` regression test. This is a generic issue and not related to this PR (see #479).

src/llvmir2hll/graphs/cfg/cfg_traversal.cpp
```diff
@@ -163,12 +176,78 @@ bool CFGTraversal::getCurrRetVal() const {
...
 		CFG::stmt_iterator stmtIter) {
-	while (stmtIter != node->stmt_end()) {
+	// The stack nodes consist of the current CFG node, a statement iterator storing the current
+	// place in the node's statements, and an iterator storing the current place in the node's
+	// successor list.
+	// Statement place is only needed because we may be called with a statement iterator
+	// that doesn't point to the beginning.
+	using NodeType = std::tuple<ShPtr<CFG::Node>, CFG::stmt_iterator, CFG::succ_iterator>;
+	std::stack<NodeType> visitStack;
+	visitStack.push({node, stmtIter, node->succ_begin()});
```
s3rvac: The Linux TravisCI build [has failed](https://travis-ci.org/avast-tl/retdec/jobs/490297897#L2276) on this line:
```
src/llvmir2hll/graphs/cfg/cfg_traversal.cpp: In member function ‘bool retdec::llvmir2hll::CFGTraversal::performTraversalImpl(retdec::llvmir2hll::ShPtr<retdec::llvmir2hll::CFG::Node>, retdec::llvmir2hll::CFG::stmt_iterator)’:
src/llvmir2hll/graphs/cfg/cfg_traversal.cpp:191:54: error: converting to ‘std::stack<...>::value_type’ from initializer list would use explicit constructor ‘constexpr std::tuple<>::tuple(...) [...]’
```
Could you please verify? It uses GCC 4.9, which is the oldest GCC version that we support. If you are unable to reproduce this, let me know and I will take a look.

s3rvac: We could simplify
```cpp
visitStack.push({node, stmtIter, node->succ_begin()});
```
to
```cpp
visitStack.emplace(node, stmtIter, node->succ_begin());
```
(Emplacement allows us to get rid of the curly brackets.)

src/llvmir2hll/graphs/cfg/cfg_traversal.cpp
```diff
@@ -163,12 +176,78 @@ bool CFGTraversal::getCurrRetVal() const {
...
+		while (succPlace != node->succ_end()) {
+			auto succNode((*succPlace)->getDst());
+			// This mutates the actual in-stack succ iterator
+			++succPlace;
+			// Push this node on the stack to be visited if it
+			// hasn't been visited and isn't empty.
+			if (succNode->stmt_begin() != succNode->stmt_end() &&
+			    !hasItem(checkedStmts, *(succNode->stmt_begin()))) {
+				pushedElement = true;
+				visitStack.push({succNode, succNode->stmt_begin(),
```
s3rvac: Similarly as in one of the above comments, consider calling `emplace()` instead of `push()` here.

src/llvmir2hll/graphs/cfg/cfg_traversal.cpp
```diff
@@ -4,6 +4,8 @@
 * @copyright (c) 2017 Avast Software, licensed under the MIT license
 */
 
+#include <stack>
+#include <tuple>
```
s3rvac: Nitpick: We split system and user includes with an empty line (see [our coding guidelines](https://github.com/avast-tl/retdec/wiki/Contribution-Guidelines#c)).

src/llvmir2hll/graphs/cfg/cfg_traversal.cpp
```diff
@@ -163,12 +176,78 @@ bool CFGTraversal::getCurrRetVal() const {
...
+	// This logic is a bit complicated because it's normally part of a depth first iterator.
+	auto retVal = getEndRetVal();
+	do {
+		// Note that we directly mutate the successor iterators that are on the
+		// stack, because we can't pop it until we are done.
+		auto node = std::get<0>(visitStack.top());
+		auto stmtPlace = std::get<1>(visitStack.top());
+		auto &succPlace = std::get<2>(visitStack.top());
+
+
```
s3rvac: Nitpick: Redundant empty line.

src/llvmir2hll/graphs/cfg/cfg_traversal.cpp
```diff
@@ -163,12 +176,78 @@ bool CFGTraversal::getCurrRetVal() const {
...
+}
+
+/**
+* @brief Visit a single node during our traversal, and all the statements in it.
+*
+* @param[in] stmtIter Iterator to a statement in node to be checked.
+* @param[in] endStmt Iterator to last statement in node to be checked.
+*
+*/
+
```
s3rvac: Nitpick: Redundant empty line.

src/llvmir2hll/graphs/cfg/cfg_traversal.cpp
```diff
@@ -163,12 +176,78 @@ bool CFGTraversal::getCurrRetVal() const {
...
+		if (stopTraversal) {
+			break;
+		}
+		// Now push the next successor on the stack to make it get
+		// visited.  We have to track whether we push an element because
+		// they may have all been visited already.
+		bool pushedElement = false;
+		while (succPlace != node->succ_end()) {
+			auto succNode((*succPlace)->getDst());
+			// This mutates the actual in-stack succ iterator
```
s3rvac: Nitpick: Missing period (`.`) at the end of the comment.

src/llvmir2hll/graphs/cfg/cfg_traversal.cpp
s3rvac: I was able to reproduce this fail in a debian-jessie Docker container. Switching from `push()` to `emplace()` and removing the curly brackets (see my other comments) resolves the compilation issue on GCC 4.9, so this is what I suggest doing.

src/llvmir2hll/graphs/cfg/cfg_traversal.cpp
s3rvac: In `master`:
```
retdec-regression-tests-framework $ ./runner.py integration.factorial -r factorial.thumb.gcc.O2.g.elf
Running 1 test case in retdec-regression-tests/integration/factorial for commit 5fb7c9fa...

integration.factorial.Test_2017 (factorial.thumb.gcc.O2.g.elf)                                      [  OK  ]  (0.90s)

SUCCESS (1/1)
```

After the changes in the PR:
```
retdec-regression-tests-framework $ ./runner.py integration.factorial -r factorial.thumb.gcc.O2.g.elf
Running 1 test case in retdec-regression-tests/integration/factorial for commit 09c7bbe0...

integration.factorial.Test_2017 (factorial.thumb.gcc.O2.g.elf)                                      [ FAIL ]  (0.89s)
F
======================================================================
FAIL: test_produce_expected_output (integration.factorial.Test_2017)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "retdec-regression-tests/integration/factorial/test.py", line 8, in test_produce_expected_output
	expected_output='factorial( 5 ) = 120\n'
  File "retdec-regression-tests-framework/regression_tests/tools/decompiler_test.py", line 87, in assert_c_produces_output_when_run
	self.assertEqual(actual_output, expected_output, msg)
AssertionError: 'factorial( 120 ) = 120\n' != 'factorial( 5 ) = 120\n'
- factorial( 120 ) = 120
?            ^^^
+ factorial( 5 ) = 120
?            ^
 : where the input was: 5

----------------------------------------------------------------------
Ran 1 test in 0.077s

FAILED (failures=1)


FAIL (failed 1 out of 1 test)
```

Here is a diff of the generated C codes (the left one is in `master`, the right one is in this PR):

diff

This might be either a bug in the new CFG traversal or in one of the optimizers that use the CFG traversal (which was triggered by a slightly different traversal order).

You can reproduce the decompilation as follows:
```
wget https://github.com/avast-tl/retdec-regression-tests/raw/master/integration/factorial/2017-11-14/factorial.thumb.gcc.O2.g.elf
python retdec-decompiler-py factorial.thumb.gcc.O2.g.elf
```

s3rvac: The difference is caused by `AuxiliaryVariablesOptimizer` (or a combination of `CopyPropagationOptimizer` and `AuxiliaryVariablesOptimizer`). It seems that `AuxiliaryVariablesOptimizer` performs an invalid optimization (`v1` is also used later in the output code, so it cannot be used to store `v5`), but I am still not sure if this is a bug in the optimizer or in the CFG traversal. I will investigate at the beginning of next week.

dberlin:
Hey Petr, i'll fix the compile failure on 4.9.

As for the regtests, how do i run those?
I have -DRETDEC_TESTS=On with cmake (i always test stuff :P), but the llvm2hir tests all pass.

If there is a way i can run the failing tests, i'm happy to debug it.

If there's no way to run it, i'll start the simple way by printing out the CFG traversal nodes on large testcases and seeing if they differ.

dberlin:
> Hey Petr, i'll fix the compile failure on 4.9.
> 
> As for the regtests, how do i run those?
> I have -DRETDEC_TESTS=On with cmake (i always test stuff :P), but the llvm2hir tests all pass.
> 
> If there is a way i can run the failing tests, i'm happy to debug it.

Forget it, found the other regression test repo.
I'll try to get it set up and running.

s3rvac:
Great :+1: Yes, all instructions are [here](https://github.com/avast-tl/retdec-regression-tests-framework#requirements). Let me know if you have any issues with the setup or execution of the tests. You can then run one of the failing tests via
```
python runner.py integration.factorial -r factorial.thumb.gcc.O2.g.elf
```
The outputs from the test will then be in
```
retdec-regression-tests/integration/factorial/outputs/Test_2017 (factorial.thumb.gcc.O2.g.elf)
```

dberlin:
I know what's wrong. I mixed too much logic into the iterator version.
Fix coming.

dberlin:
I've updated it to the latest version of the patch, which just moves the logic into a (local) depth first iterator to make it easier to debug.  

I have not finished making all requested changes, but wanted to put a working version here.  It passes the regression test the other one failed.

I've tested this version by instrumenting pre/post versions of the compiler to print traversals and ensuring the same traversals occur in the same order (using https://gist.github.com/39013640d89f08ec285a04e68d7197bf for the pre).

There are a few spurious differences in traversals i am tracking down  , but they don't affect generated code AFAICT.  

Some of them are caused by the fact that functions in compinfo/et al  are sorted in pointer order so the overall processing order of optimizations is not deterministic.

(I know y'all do SCC finding, and it definitely gives a topo order of SCC's overall, but the order in which the SCCs and functions appear in the sets is based on pointer address. Even with disabling ASLR, they are not identical from run to run.  This is easy to see by printing the function names as they are processed).

s3rvac:
Thank you for the fix :+1: As for the non-determinism in our tools, we are aware of that (#209, #479) and we would like to gradually fix it. Several of the collections that we use should have been vectors, but we realized that long after we had used sets in the implementation.

s3rvac:
Hi Daniel. We are planning to release a new version of RetDec. It would be cool if this PR could be part of that release. Can you please summarize the current status of this PR?

dberlin:
Sure. I tested the updated version before I had to run away and it seems to
work fine for regression tests/etc.  It is, AFAICT, functionally correct at
this point.  I haven't had a chance to see if there are comment/other
requested changes. and unfortunately i don't think i'll get back to it for
a while :(
(I had a bunch of free time and now i suddenly don't)

s3rvac:
That's alright. All our tests pass. Is it OK if I merge the branch into `master`?

dberlin:
It is perfectly fine by me!

src/llvmir2hll/graphs/cfg/cfg_traversal.cpp
s3rvac: All tests pass now, so I am marking this as resolved.

src/llvmir2hll/graphs/cfg/cfg_traversal.cpp
s3rvac: The TravisCI build runs just fine, so I am marking this as resolved.

s3rvac:
Great. Thank you for the valuable contribution! :+1:

PR #499, opened 2019-02-14
Change SCC's to be FuncVectorSet instead of FuncSetSet.

dberlin:
Previously, the set of sccs was sorted by overall function pointer order.
This was one source of non-determinism.
It also  destroyed the topological ordering of SCCs that tarjan creates.

Also, insertion of each funcset (scc) into the tree of funcsetset possibly compared it against log N funcsets, where each comparison is O(N) (The definition of operator< on sets
performs lexicographic compare on all the members of each set).
This did not do well in the face when inserting large sccs since it may walk a many-thousand member set log N times just to determine the funcset's place in the funcsetset tree.

In practice this is a minor time savings but given we don't want them sorted this way anyway ...

### DIFF
--- a/include/retdec/llvmir2hll/obtainer/call_info_obtainer.h
+++ b/include/retdec/llvmir2hll/obtainer/call_info_obtainer.h
@@ -251,8 +251,8 @@ public:
 	virtual ShPtr<FuncInfo> getFuncInfo(ShPtr<Function> func) = 0;
 
 protected:
-	/// Set of sets of functions.
-	using FuncSetSet = std::set<FuncSet>;
+	/// Vector of sets of functions.
+	using FuncVectorSet = std::vector<FuncSet>;
 
 	/**
 	* @brief Represents an order in which FuncInfos should be computed.
@@ -297,7 +297,7 @@ protected:
 		FuncVector order;
 
 		/// SCCs in the call graph.
-		FuncSetSet sccs;
+		FuncVectorSet sccs;
 	};
 
 	/// Mapping of a function into its CFG.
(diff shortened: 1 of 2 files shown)


src/llvmir2hll/obtainer/call_info_obtainer.cpp
```diff
@@ -436,7 +437,7 @@ void CallInfoObtainer::SCCComputer::visit(ShPtr<CG::CalledFuncs> calledFunc,
 		// function, do this only if it calls itself (see the description of
 		// computeSCCs()).
 		if (scc.size() != 1 || hasItem(calledFunc->callees, calledFunc->caller)) {
-			sccs.insert(scc);
+			sccs.push_back(scc);
```
s3rvac: Here, we could mention that Tarjan's algorithm never creates the same SCC twice, so it is safe to use just `push_back()` without any existence checks. What do you think?

include/retdec/llvmir2hll/obtainer/call_info_obtainer.h
```diff
@@ -252,7 +252,7 @@ class CallInfoObtainer: public SharableFromThis<CallInfoObtainer>,
 
 protected:
 	/// Set of sets of functions.
```
s3rvac: Is this description correct? I am referring to _"Set of sets"_. Technically, it is no longer a set. However, from a high-level point of view, I can understand the the description might be OK. What is your take on this?

We should also maybe consider adding a comment why a vector is a better choice here than a set.

src/llvmir2hll/obtainer/call_info_obtainer.cpp
dberlin: Yup. I'm going to probably rewrite this code a bit later, but it's always good to have accurate comments at all times. Fixed

include/retdec/llvmir2hll/obtainer/call_info_obtainer.h
dberlin: Let's just go with vector.
I've fixed this comment and added a comment to the computeSCCs explaining why we use it.

PR #440, opened 2018-12-13
Visual Basic Information Parsing, PCode detection

JakubPruzinec:
To be completed.

PR #501, opened 2019-02-14
Computation of section entropies for ELF, MACHO and PE (in progress)

JakubPruzinec:
https://github.com/avast-tl/retdec/issues/502
I've implemented extraction of section entropies for various file formats.
Tests are not implemented yet, thus the "in progress" status.


s3rvac:
It seems that this branch is based on the branch for #440 (it contains the same commits). Shouldn't it be based on `master`?

JakubPruzinec:
right, i'll fix that

src/fileformat/file_format/file_format.cpp
```diff
@@ -1091,6 +1091,24 @@ std::size_t FileFormat::getOverlaySize() const
...
+/**
+ * Get overlay data entropy
+ * @param res Variable to store the result to
+ * @return @c true if entropy calculation succeeded, @false otherwise
+ */
+bool FileFormat::getOverlayEntropy(double &res) const
+{
+	const auto overlaySize = getOverlaySize();
+	const auto declSize = getDeclaredFileLength();
+	const auto bytes = getBytes();
```
s3rvac: Beware that storing the result of
```cpp
const std::vector<unsigned char>& getBytes() const;
```
into
```cpp
const auto bytes = getBytes();
```
causes the vector to be copied, i.e. the type of `bytes` will be `const std::vector<unsigned char>`. I believe what you want is
```cpp
const auto &bytes = getBytes();
```
which will store a reference, i.e. the type of `bytes` will be `const std::vector<unsigned char> &`.

src/fileformat/types/sec_seg/sec_seg.cpp
```diff
@@ -521,6 +536,27 @@ void SecSeg::setMemory(bool sMemory)
...
+ * Compute entropy of section data in <0,1>
+ */
+void SecSeg::computeEntropy()
+{
+	if (!loaded)
+	{
+		return;
+	}
+	
+	auto data = reinterpret_cast<const uint8_t *>(bytes.data());
```
s3rvac: Do we need a `reinterpret_cast` cast for this? Doesn't a `static_cast` suffice?

src/fileinfo/file_detector/macho_detector.cpp
```diff
@@ -118,6 +119,10 @@ void MachODetector::getSections()
 		fsec.setMemoryAlignment(pow(2, mSec->getAlignment()));
 		fsec.setRelocationsOffset(mSec->getRelocationOffset());
 		fsec.setNumberOfRelocations(mSec->getNumberOfRelocations());
+		if(sec->getEntropy(entropy))
```
s3rvac: I suggest defining `double entropy;` right before `if(sec->getEntropy(entropy))`, not 14 lines before. It is generally a good practice to define variables as close to their first use, especially when they are not initialized.

src/fileformat/types/sec_seg/sec_seg.cpp
JakubPruzinec: I belive not https://stackoverflow.com/questions/658913/c-style-cast-from-unsigned-char-to-const-char

PR #507, opened 2019-02-23
Entropy calculation of sections and overlays (PE, ELF, COFF, MACHO)

JakubPruzinec:
https://github.com/avast-tl/retdec/issues/502
I've implemented extraction of section and overlay entropies. Algorithm for entropy calculation uses Shannons formula normalized to ~<0,1>~ <0,8> interval.


src/fileformat/types/sec_seg/sec_seg.cpp
```diff
@@ -521,6 +536,27 @@ void SecSeg::setMemory(bool sMemory)
...
+ * Compute entropy of section data in <0,1>
+ */
+void SecSeg::computeEntropy()
+{
+	if (!loaded)
+	{
+		return;
+	}
+	
+	auto data = reinterpret_cast<const uint8_t *>(bytes.data());
```
metthal: Is `reinterpret_cast` really necessary here? Wouldn't `static_cast` be enough or is the cast necessary at all?

src/fileinfo/file_wrapper/pe/pe_wrapper.cpp
```diff
@@ -176,9 +176,15 @@ bool PeWrapper::getFileSection(unsigned long long secIndex, FileSection &section
...
 		{
+			double entropy;
+			if(auxSec->getEntropy(entropy))
+			{
+				section.setEntropy(entropy);
+			}
 			section.setCrc32(auxSec->getCrc32());
 			section.setMd5(auxSec->getMd5());
 			section.setSha256(auxSec->getSha256());
+
```
metthal: Remove this unnecessary whitespace please.

src/fileformat/types/sec_seg/sec_seg.cpp
JakubPruzinec: Cast is necessary, since `bytes.data()` returns `const char *` and `computeDataEntropy` takes `const unsigned char*` as parameter. As stated earlier, `static_cast` doesn't suffice. `invalid static_cast from type ‘const char*’ to type ‘const uint8_t*’ {aka ‘const unsigned char*’}`, please see https://stackoverflow.com/questions/658913/c-style-cast-from-unsigned-char-to-const-char

JakubPruzinec: Also, I wouln't be so concerned about wrong interpretation of signedness as `computeDataEntropy` doesn't take it into consideration.

metthal:
Could you please also modify it so that results are in range from 0 to 8? We should stick with what [VirusTotal](https://www.virustotal.com/#/file/1054ab5902dd121ab49ae855415926bbc5b5ed974444cc8543a7b9388a82a098/details) and [YARA](https://github.com/VirusTotal/yara/blob/master/libyara/modules/math.c#L68) are using.

src/fileformat/utils/other.cpp
```diff
@@ -99,5 +101,38 @@ std::vector<std::string> getSupportedArchitectures()
...
+	{
+		histogram[data[i]]++;
+	}
+
+	for (auto frequency : histogram)
+	{
+		if (frequency)
+		{
+			double probability = static_cast<double>(frequency) / dataLen;
+			entropy -= probability * (log(probability) / log(2));
```
metthal: This can be calculated using only single logarithm because `log_a(x) / log_a(b) = log_b(x)` so you can actually use logarithm of base-2 here. Also please use `std::` namespace instead of the functions in the global namespace.

src/fileformat/utils/other.cpp
```diff
@@ -4,8 +4,10 @@
  * @copyright (c) 2017 Avast Software, licensed under the MIT license
  */
 
+#include <array>
 #include <map>
 #include <unordered_map>
+#include <cmath>
```
metthal: Please keep headers in alphabetical order.

include/retdec/fileformat/types/sec_seg/sec_seg.h
```diff
@@ -47,10 +47,12 @@ class SecSeg
 		unsigned long long address;       ///< start address in memory
 		unsigned long long memorySize;    ///< size in memory
 		unsigned long long entrySize;     ///< size of one entry in file
+		double entropy;                   ///< entropy in <0,1>
```
metthal: Isn't comment here wrong? Shouldn't it be `<0,8>`?

PR #509, opened 2019-02-26
Remove the original converter of LLVM IR into BIR from llvmir2hll

### LINKED ISSUE #508: Remove the original converter of LLVM IR into BIR from llvmir2hll
Now that `retdec-llvmir2hll` is using the new converter of LLVM IR into BIR, we can remove the original (old) one. Reasons:
* Nobody will probably ever use it.
* We will not have to maintain two converters.
* Fewer `.cpp` files to compile.

s3rvac:
Now that retdec-llvmir2hll is using the new converter of LLVM IR into BIR, we can remove the original (old) one.

Reasons:
* Nobody will probably ever use it.
* We will not have to maintain two converters.
* Fewer `.cpp` files to compile.

Notes:
* We have to preserve `labels_handler.cpp` from the original converter as this module is also used by the new converter.
* The `retdec-decompiler.py` script had to be also updated (to prevent users from specifying `'orig'` via `--backend-llvmir2bir-converter`.

Closes #508.


PeterMatula:
Does it make sense to still call the converter `new` if it is the only one left? I.e. what about changing source file names, comments, etc.
The question is, if the architecture allows to have more than one converter, and we remove `new` from its name, how will we ID it? In case there is ever another converter.

s3rvac:
Good point. I see two options:
1. Rename the converter to e.g. `current` and keep the possibility of having more converters.
2. Remove support for having more converters and keep just a single one (without a name).

I would go with number 2 as I do not believe we will have another converter in the future. Also, having multiple converters would be unnecessarily hard to maintain (we would have to write regression tests for both of them).

What is your take on this?

s3rvac:
Based on a discussion, we have decided to go with option number 2. I will adjust the PR to reflect that.

s3rvac:
Done in commit 97594279ad3bb99acca8919527facbe3ccb00787.

Issue #458, opened 2019-01-07
fileinfo crashes during reconstruction of .NET types


s3rvac:
`fileinfo` crashes during reconstruction of .NET types in the PE binary below.

## Input

Run
```
$ retdec-fileinfo FILE
```
where `FILE` is:
* [`057CC3829D8EBA8AEBC3043EEA40AF5B05E1EB229012136B43134728E1B46D63`](https://www.virustotal.com/#/file/057cc3829d8eba8aebc3043eea40af5b05e1eb229012136b43134728e1b46d63/detection)
* [`54FB1B3D36EB3FB13A3C565A7C7A0A7E2190E0724014488DF60713F2276A379C`](https://www.virustotal.com/#/file/54fb1b3d36eb3fb13a3c565a7c7a0a7e2190e0724014488df60713f2276a379c/detection)

## Output

```
Segmentation fault
```

## Expected output

`fileinfo` does not crash when analyzing the file.

## Output from valgrind

```
==16969== Invalid read of size 8
==16969==    at 0x2DBA3C: std::vector<retdec::fileformat::AssemblyRef, std::allocator<retdec::fileformat::AssemblyRef> >::size() const (stl_vector.h:806)
==16969==    by 0x2D835E: retdec::fileformat::MetadataTable<retdec::fileformat::AssemblyRef>::getRow(unsigned long) const (metadata_table.h:98)
==16969==    by 0x382320: retdec::fileformat::DotnetTypeReconstructor::createClassReference(retdec::fileformat::TypeRef const*, unsigned long) (dotnet_type_reconstructor.cpp:843)
==16969==    by 0x3802D1: retdec::fileformat::DotnetTypeReconstructor::reconstructClasses() (dotnet_type_reconstructor.cpp:401)
==16969==    by 0x37F853: retdec::fileformat::DotnetTypeReconstructor::reconstruct() (dotnet_type_reconstructor.cpp:227)
==16969==    by 0x2C588B: retdec::fileformat::PeFormat::detectDotnetTypes() (pe_format.cpp:1895)
==16969==    by 0x2C3151: retdec::fileformat::PeFormat::loadDotnetHeaders() (pe_format.cpp:1259)
==16969==    by 0x2BE54D: retdec::fileformat::PeFormat::initStructures() (pe_format.cpp:403)
==16969==    by 0x2BDC93: retdec::fileformat::PeFormat::PeFormat(...) (pe_format.cpp:297)
[...]

PR #511, opened 2019-02-26
TypeRef table reconstruction SEGFAULT fix

JakubPruzinec:
https://github.com/avast-tl/retdec/issues/458
I've added a check for null in TypeRef table reconstruction to prevenet null dereference.

### DIFF
--- a/src/fileformat/types/dotnet_types/dotnet_type_reconstructor.cpp
+++ b/src/fileformat/types/dotnet_types/dotnet_type_reconstructor.cpp
@@ -840,6 +840,12 @@ std::unique_ptr<DotnetClass> DotnetTypeReconstructor::createClassReference(const
 	else
 	{
 		auto assemblyRefTable = static_cast<const MetadataTable<AssemblyRef>*>(metadataStream->getMetadataTable(MetadataTableType::AssemblyRef));
+
+		if (!assemblyRefTable)
+		{
+			return nullptr;
+		}
+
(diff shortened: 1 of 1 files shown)


src/fileformat/types/dotnet_types/dotnet_type_reconstructor.cpp
```diff
@@ -840,6 +840,12 @@ std::unique_ptr<DotnetClass> DotnetTypeReconstructor::createClassReference(const
 	else
 	{
 		auto assemblyRefTable = static_cast<const MetadataTable<AssemblyRef>*>(metadataStream->getMetadataTable(MetadataTableType::AssemblyRef));
+
+		if (!assemblyRefTable)
```
s3rvac: Would it be possible/desirable to create a regression test for `fileinfo` from of the input binaries from #458? For example from the smaller one.

metthal:
Thanks for fix. LGTM but add regression test for the smallest binary from #458 and that's all.

src/fileformat/types/dotnet_types/dotnet_type_reconstructor.cpp
JakubPruzinec: Sure, unfortunately the smaller binary contains no parsable `TypeRefTable`, therefore i made RT with the bigger one. https://github.com/avast-tl/retdec-regression-tests/pull/19

PR #523, opened 2019-03-19
TLS parsing and presentation

JakubPruzinec:
Thread-local storage information and callback extraction
https://github.com/avast-tl/retdec/issues/417

### DIFF
--- a/include/retdec/fileformat/file_format/pe/pe_format_parser/pe_format_parser64.h
+++ b/include/retdec/fileformat/file_format/pe/pe_format_parser/pe_format_parser64.h
@@ -68,6 +68,12 @@ class PeFormatParser64 : public PeFormatParser
 		virtual bool getDebugEntryPointerToRawData(unsigned long long index, unsigned long long& pointerToRawData) const override;
 		virtual unsigned long long getResourceDirectoryOffset() const override;
 		virtual const PeLib::ResourceNode* getResourceTreeRoot() const override;
+		virtual unsigned long long getTlsStartAddressOfRawData() const;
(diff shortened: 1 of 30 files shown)


include/retdec/fileformat/file_format/pe/pe_format_parser/pe_format_parser64.h
```diff
@@ -68,6 +68,12 @@ class PeFormatParser64 : public PeFormatParser
 		virtual bool getDebugEntryPointerToRawData(unsigned long long index, unsigned long long& pointerToRawData) const override;
 		virtual unsigned long long getResourceDirectoryOffset() const override;
 		virtual const PeLib::ResourceNode* getResourceTreeRoot() const override;
+		virtual unsigned long long getTlsStartAddressOfRawData() const;
```
metthal: `override` is missing here and also on other lines under this one.

metthal:
I went through the code once more and I am not sure whether this will correctly work for all 64-bit PE binaries. PeLib seems to be truncating half of bytes when returning attributes which are 64-bit. I think this may require modification of PeLib, more specifically `TlsDirectory` in PeLib and turning all attributes which are either 32-bit/64-bit depending on the PE32/PE32+ into `typename FieldSizes<bits>::VAR4_8`. Those are:

```
StartAddressOfRawData
EndAddressOfRawData
AddressOfIndex
AddressOfCallBacks
```

metthal:
Thanks for taking care of PE32+ but it will still need modification of pelib because it truncates it to 32-bit values (`dword`).

PR #533, opened 2019-03-26
Aarch64 decompilation

MatejKastak:
Changes include instructions translation to llvmir, unit tests, llvmir environment generation, arch abi, etc.


PeterMatula:
Although I said that you don't need to document every single thing in Doxygen, please make sure that the existing comments are without errors - `doxygen-build` doesn't fail.

MatejKastak:
Yes, I completely forgot about documentation builds. It should be fixed now.

PR #543, opened 2019-04-11
fileformat: Hashes of certificates changed to lowercase

stepanek-m:
Hashes in certificate table changed to lowercase in order to be consistent with other hashes in fileinfo.

### DIFF
--- a/src/fileformat/types/certificate_table/certificate.cpp
+++ b/src/fileformat/types/certificate_table/certificate.cpp
@@ -244,7 +244,7 @@ void Certificate::loadSignatureAlgorithm()
 void Certificate::loadSerialNumber()
 {
 	if (auto sn = X509_get_serialNumber(certImpl))
-		retdec::utils::bytesToHexString(sn->data, sn->length, serialNumber, 0, 0, false);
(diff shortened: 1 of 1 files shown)


src/fileformat/types/certificate_table/certificate.cpp
```diff
@@ -244,7 +244,7 @@ void Certificate::loadSignatureAlgorithm()
 void Certificate::loadSerialNumber()
 {
 	if (auto sn = X509_get_serialNumber(certImpl))
-		retdec::utils::bytesToHexString(sn->data, sn->length, serialNumber);
+		retdec::utils::bytesToHexString(sn->data, sn->length, serialNumber, 0, 0, false);
```
s3rvac: Consider changing
```
retdec::utils::bytesToHexString(sn->data, sn->length, serialNumber, 0, 0, false);
```
to
```
retdec::utils::bytesToHexString(sn->data, sn->length, serialNumber, 0, 0, /*uppercase=*/false);
                                                                          ^^^^^^^^^^^^^^
```
so the meaning of the `false` argument is clear. Unfortunately, C++ does not have something akin to keyword arguments from Python :(.

Issue #415, opened 2018-10-15
fileinfo: Anomalies in PE files


metthal:
PE files (or generally other executable files, but let's stick to PE files in this ticket) do not necessarily need to conform to the PE specification in order to be loadable by Windows. There are often many discrepancies  and violations of the specification and the file still remains valid. This is often abused by malware authors in order to make analysis of the file more complicated. These, so called _anomalies_, can be used as initial indicators that something fishy is going on. They can also be used by another tools which use `fileinfo` as their source of data in order to focus on things pointed out by these indicators.

The indicators would be presented by `fileinfo` in form of list. Indicators would also possibly be parameterized if it is necessary. Indicators can also possibly be split into categories by their severity. 

I am not going to list here all the possible anomalies that are going to be implemented here because I don't even know them all myself and the list would be infinite. We can start with the detection of unusual values in PE header and work our way from that. This is more of a research that definite task.

PeterMatula:
Internal note: possible inspiration in Avast cuckoo module `static_pe_anomaly.py`.

metthal:
Additional anomalies proposed:
* Detection of PDB paths that was purposefully truncated (like in `af4db267da79adc536f4fdae30860d92b384a39acb8a7fa6ed1b99ad66310188`)

PR #570, opened 2019-05-06
Anomalies scanner

JakubPruzinec:
https://github.com/avast/retdec/issues/415
I've implemented PE scanner for format anomalies.

PE format is scanned by:
* scanForSectionAnomalies()
* scanForResourceAnomalies()
* scanForImportAnomalies()
* scanForExportAnomalies()
* scanForOptHeaderAnomalies()
* scanForObsoleteCharacteristics()


src/fileformat/file_format/pe/pe_format.cpp
```diff
@@ -3350,5 +3409,320 @@ const VisualBasicInfo* PeFormat::getVisualBasicInfo() const
...
+	std::size_t nSecs = getDeclaredNumberOfSections();
+	
+	const PeCoffSection *epSec = dynamic_cast<const PeCoffSection*>(getEpSection());
+	if (epSec)
+	{
+		// scan EP in last section
+		const PeCoffSection *lastSec = (nSecs) ? getPeSection(nSecs - 1) : nullptr;
+		if (epSec == lastSec)
+		{
+			anomalies.emplace_back(std::make_pair<std::string, std::string>("epInLastSec",
```
s3rvac: Two remarks concerning the use of `std::make_pair()` and `emplace_back()`:
1. When using `std::make_pair`, you are not supposed to specify the template parameters (see e.g. [this SO answer](https://stackoverflow.com/a/9642232/2580955)). The whole point of `std::make_pair()` is that you don't have to specify the template parameters. Otherwise, you could just as well create `std::pair` directly.
2. There is no benefit of using `emplace_back()` with r-values (you could call just `push_back()`). The benefit of `emplace_back()` comes from the fact that you can pass just the arguments to the constructor of the objects that a vector holds, which both simplifies the code and possibly makes it faster, as the object is created in-place.

Thus, instead of
```cpp
anomalies.emplace_back(std::make_pair<std::string, std::string>("x", "y"));
```
write just
```cpp
anomalies.emplace_back("x", "y");
```

src/fileformat/file_format/pe/pe_format.cpp
```diff
@@ -3350,5 +3409,320 @@ const VisualBasicInfo* PeFormat::getVisualBasicInfo() const
...
+	}
+}
+
+/**
+ * Scan for import anomalies
+ */
+void PeFormat::scanForImportAnomalies()
+{
+	// scan for import stretched over multiple sections
+	for(auto&& impRange : formatParser->getImportDirectoryOccupiedAddresses())
```
s3rvac: Do we really need a forwarding reference in this case? Wouldn't `const auto&` suffice?
```cpp
for(const auto& impRange : formatParser->getImportDirectoryOccupiedAddresses())
```
As Howard Hinnant puts it, liberate use of `auto&&` results in so-called [confuscated code](http://stackoverflow.com/a/13130795/2580955): code that unnecessarily confuses people.

Similar piece of code appears in `PeFormat::scanForExportAnomalies()`.

include/retdec/fileformat/file_format/pe/pe_format.h
```diff
@@ -107,6 +107,15 @@ class PeFormat : public FileFormat
 		bool parseVisualBasicComRegistrationInfo(std::size_t structureOffset,
 												std::size_t comRegDataOffset);
 		/// @}
+		/// @name Auxiliary scanning methods
+		/// @{
+		void scanForSectionAnomalies();
+		void scanForResourceAnomalies();
+		void scanForImportAnomalies();
+		void scanForExportAnomalies();
+        void scanForOptHeaderAnomalies();
```
metthal: TABs

src/fileformat/file_format/pe/pe_format.cpp
```diff
@@ -3350,5 +3409,299 @@ const VisualBasicInfo* PeFormat::getVisualBasicInfo() const
...
+		const std::string msgName = (name.empty()) ? numToStr(sec->getIndex()) : name;
+		const auto flags = sec->getPeCoffFlags();
+		if (!name.empty())
+		{
+			// scan for unusual section names
+			if (std::find(usualSectionNames.begin(), usualSectionNames.end(), name) == usualSectionNames.end())
+			{
+				if (std::find(duplSections.begin(), duplSections.end(), name) == duplSections.end())
+				{
+					anomalies.emplace_back("unusualSecName", "Unusual section name: " + name);
```
metthal: Escape section names because you might end up with gibberish output.

src/fileformat/file_format/pe/pe_format.cpp
```diff
@@ -3350,5 +3409,299 @@ const VisualBasicInfo* PeFormat::getVisualBasicInfo() const
...
+				}
+			}
+
+			// scan for packer section names
+			if (std::find(usualPackerSections.begin(), usualPackerSections.end(), name) != usualPackerSections.end())
+			{
+				if (std::find(duplSections.begin(), duplSections.end(), name) == duplSections.end())
+				{
+					anomalies.emplace_back("packedSecName", "Packer section name: " + name);
```
metthal: Escape section names because you might end up with gibberish output.

src/fileformat/file_format/pe/pe_format.cpp
```diff
@@ -3350,5 +3409,299 @@ const VisualBasicInfo* PeFormat::getVisualBasicInfo() const
...
+				}
+			}
+
+			// scan for unexpected characteristics
+			auto characIt = usualSectionCharacteristics.find(name);
+			if (characIt != usualSectionCharacteristics.end() && characIt->second != flags)
+			{
+				anomalies.emplace_back("unusualSecChar", "Section " + name + " has unusual characteristics");
```
metthal: Escape section names because you might end up with gibberish output.

src/fileformat/file_format/pe/pe_format.cpp
```diff
@@ -3350,5 +3409,299 @@ const VisualBasicInfo* PeFormat::getVisualBasicInfo() const
...
+			}
+		}
+
+		// scan size over 100MB
+		if (sec->getSizeInFile() >= 100000000UL)
+		{
+			anomalies.emplace_back("largeSec", "Section " + msgName + " has size over 100MB");
```
metthal: Escape section names because you might end up with gibberish output.

src/fileformat/file_format/pe/pe_format.cpp
```diff
@@ -3350,5 +3409,299 @@ const VisualBasicInfo* PeFormat::getVisualBasicInfo() const
...
+		}
+
+		// scan section marked uninitialized but contains data
+		if ((flags & PELIB_IMAGE_SCN_CNT_UNINITIALIZED_DATA) &&
+			(sec->getOffset() != 0 || sec->getSizeInFile() != 0))
+		{
+			anomalies.emplace_back("uninitSecHasData", "Section " + msgName + " is marked uninitialized but contains data");
```
metthal: Escape section names because you might end up with gibberish output.

src/fileformat/file_format/pe/pe_format.cpp
```diff
@@ -3350,5 +3409,299 @@ const VisualBasicInfo* PeFormat::getVisualBasicInfo() const
...
+		}
+
+		// scan sizeOfRawData of section is 0
+		if (sec->getSizeInFile() == 0)
+		{
+			anomalies.emplace_back("noRawDataSec", "Section " + msgName + " has zero SizeOfRawData");
```
metthal: Escape section names because you might end up with gibberish output.

src/fileformat/file_format/pe/pe_format.cpp
```diff
@@ -3350,5 +3409,299 @@ const VisualBasicInfo* PeFormat::getVisualBasicInfo() const
...
+				continue;
+			}
+
+			// scan for duplicit section names
+			const auto cmpName = cmpSec->getName();
+			if (!name.empty() && name == cmpName)
+			{
+				if (std::find(duplSections.begin(), duplSections.end(), name) == duplSections.end())
+				{
+					anomalies.emplace_back("duplSecNames", "Multiple sections with name " + name);
```
metthal: Escape section names because you might end up with gibberish output.

src/fileformat/file_format/pe/pe_format.cpp
```diff
@@ -3350,5 +3409,299 @@ const VisualBasicInfo* PeFormat::getVisualBasicInfo() const
...
+			// scan for overlapping sections
+			auto secStart = sec->getOffset();
+			auto secEnd = secStart + sec->getSizeInFile();
+			auto cmpSecStart = cmpSec->getOffset();
+			auto cmpSecEnd = cmpSecStart + cmpSec->getSizeInFile();
+			if ((secStart <= cmpSecStart && cmpSecStart < secEnd) ||
+				(cmpSecStart <= secStart && secStart < cmpSecEnd))
+			{
+				const std::string cmpMsgName = (cmpName.empty()) ? numToStr(cmpSec->getIndex()) : cmpName;
+				anomalies.emplace_back("overlappingSecs", "Sections " + msgName + " and " + cmpMsgName + " overlap");
```
metthal: Escape section names because you might end up with gibberish output.

src/fileformat/file_format/pe/pe_format.cpp
```diff
@@ -3350,5 +3409,299 @@ const VisualBasicInfo* PeFormat::getVisualBasicInfo() const
...
+					}
+
+				}
+				else
+				{
+					msgName = imp->getName();
+				}
+			}
+
+			anomalies.emplace_back("stretchedImp", "Import " + msgName + " is stretched over multiple sections");
```
metthal: Escape import names because you might end up with gibberish output.

src/fileformat/file_format/pe/pe_format.cpp
```diff
@@ -3350,5 +3409,299 @@ const VisualBasicInfo* PeFormat::getVisualBasicInfo() const
...
+					}
+
+				}
+				else
+				{
+					msgName = exp->getName();
+				}
+			}
+
+			anomalies.emplace_back("stretchedExp", "Export " + msgName + " is stretched over multiple sections");
```
metthal: Escape export names because you might end up with gibberish output.

src/fileformat/file_format/pe/pe_format.cpp
```diff
@@ -3350,5 +3409,299 @@ const VisualBasicInfo* PeFormat::getVisualBasicInfo() const
...
+	}
+}
+
+/**
+ * Scan for obsolete file characteristics
+ */
+void PeFormat::scanForObsoleteCharacteristics()
+{
+	std::size_t flags = getFileFlags();	
+	std::size_t obsoleteFlags = PELIB_IMAGE_FILE_LINE_NUMS_STRIPPED |
```
metthal: Where did you take this anomaly from? I am just curious whether each of this flag is really obsolete. https://docs.microsoft.com/en-us/windows/desktop/api/winnt/ns-winnt-_image_file_header says only few of the ones you 've listed are.

JakubPruzinec: I have adopted it directly from source codes of [PortEx](https://github.com/katjahahn/PortEx). The link you posted seems to ommit the fact that the flags are indeed obsolete (see [flags documentation](https://docs.microsoft.com/en-us/windows/desktop/debug/pe-format#characteristics) referenced by [PE documentation](https://docs.microsoft.com/en-us/windows/desktop/debug/pe-format#coff-file-header-object-and-image) ). I checked whether the `Obsolete Flag` anomaly pops on binary I have created with VB6 compiler and it does. Consider neglecting the flags that turned obsolete "recently".

PR #586, opened 2019-06-10
Properly escape backslash in escaped strings

metthal:
This fixes problems with string that have escaped non-printable chars but do not have escaped backslash. When parsing such string, you are not able correctly reconstruct it because it is ambiguous.


metthal:
Closing the pull request since we've decided we are satisfied with our current solution. The proposed solution caused that in the plain output, all backslashes were doubled and it became non-readable and was not matching with the output of other tools.

PR #588, opened 2019-06-18
make flag clarification

sveip:
N -j flag denotes parallel processes

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,7 +236,7 @@ Note: Although RetDec now supports a system-wide installation ([#94](https://git
     sudo pkg install cmake python37 bison autotools
     ```
   * `cmake .. -DCMAKE_INSTALL_PREFIX=<path>`
-  * `make -jN` (`N` is the number of  of processes to use for parallel build, typically number of cores + 1 gives fastest compilation time)
+  * `make -jN` (`N` is the number of processes to use for parallel build, typically number of cores + 1 gives fastest compilation time)
   * `make install`
 
 You have to pass the following parameters to `cmake`:
--- a/README.md
+++ b/README.md
@@ -200,7 +200,7 @@ Note: Although RetDec now supports a system-wide installation ([#94](https://git
   * `cd retdec`
   * `mkdir build && cd build`
   * `cmake .. -DCMAKE_INSTALL_PREFIX=<path>`
-  * `make -jN` (`N` is the number of CPU cores to use for parallel build)
+  * `make -jN` (`N` is the number of  of processes to use for parallel build, typically number of cores + 1 gives fastest compilation time)
   * `make install`
 * Windows:
   * Open a command prompt (e.g. `cmd.exe`)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -200,7 +200,7 @@ Note: Although RetDec now supports a system-wide installation ([#94](https://git
   * `cd retdec`
   * `mkdir build && cd build`
   * `cmake .. -DCMAKE_INSTALL_PREFIX=<path>`
-  * `make -jN` (`N` is the number of CPU cores to use for parallel build)
+  * `make -jN` (`N` is the number of  of processes to use for parallel build, typically number of cores + 1 gives fastest compilation time)
```
s3rvac: In all cases, there is `of  of processes` (double `of` and double space) instead of `of processes`.

README.md
```diff
@@ -220,7 +220,7 @@ Note: Although RetDec now supports a system-wide installation ([#94](https://git
     export PATH="/usr/local/opt/flex/bin:/usr/local/opt/bison/bin:$PATH"
     ```
   * `cmake .. -DCMAKE_INSTALL_PREFIX=<path>`
-  * `make -jN` (`N` is the number of CPU cores to use for parallel build)
+  * `make -jN` (`N` is the number of  of processes to use for parallel build, typically number of cores + 1 gives fastest compilation time)
```
s3rvac: Is there some evidence behind `typically number of cores + 1 gives fastest compilation time` with relation to RetDec? I have just tried that with RetDec and that +1 made the build actually slightly slower.

I have seen this advice a few times in the past, but the results seem to vary greatly in practice (sometimes the build is faster, sometimes slower). Often, using the number of cores + 1 actually makes the system overloaded. So, I would go with `; typically, the number of cores gives fastest compilation time`. What do you think? This is just a nitpick, so feel free to keep your version if you believe it is better.

sveip: I think the reson behind this it that if you have one extra process, you can still have one of them stalled on I/O and still utilize your CPU to the max. With fast SSDs today, it might not be such a big deal. About stall, in my experience you would have to increase the number to something much larger than ¤ CPU cores to make the system stall.

s3rvac: Fair enough. Feel free to keep the current wording.

README.md
```diff
@@ -236,7 +236,7 @@ Note: Although RetDec now supports a system-wide installation ([#94](https://git
     sudo pkg install cmake python37 bison autotools
     ```
   * `cmake .. -DCMAKE_INSTALL_PREFIX=<path>`
-  * `make -jN` (`N` is the number of CPU cores to use for parallel build)
+  * `make -jN` (`N` is the number of  of processes to use for parallel build, typically number of cores + 1 gives fastest compilation time)
```
s3rvac: Nitpick: One missed occurrence of `of  of`. Can you please also fix it before I merge this?

sveip: Sorry, I was lazy and used the online github editor, I have learned my lesson :)

PR #614, opened 2019-07-22
Pass info of used calling conventions into config

xkubov:
Hi, this PR brings minor modifications into parameter optimization. With these modifications is info about calling conventions (that were used during the parameter analysis) projected into a configuration JSON output file after the optimization.

All architectures with only one calling convention set name of the calling convention to the config as `"arch_default"` (eg. `arm_default`, `mips_default`, etc.).

When it comes to x64, this architecture has two calling conventions that are used on different operating systems. On Windows, calling convention that is used is Microsoft x64. On UNIX like systems, it is System V convention.
For simplicity is therefor name of the convention saved into the config file as `"x64_os_default"`.

Example of `config.json` on x86:
```
{  
   "callingConvention":"cdecl",
   "name":"main",
   "parameters": [  
...
},
```

and before changes:
```
{  
   "callingConvention":"unknown",
   "name":"main",
   "parameters": [  
...
},
```

### DIFF
--- a/src/config/calling_convention.cpp
+++ b/src/config/calling_convention.cpp
@@ -24,7 +24,16 @@ const std::vector<std::string> ccStrings =
 	"spoiled",
 	"speciale",
 	"specialp",
-	"special"
+	"special",
+	"watcom",
+	"x64_os_default",
+	"arm_default",
+	"arm64_default",
+	"mips_default",
+	"mips64_default",
+	"powerpc_default",
+	"powerpc64_default",
+	"pic32_default"
 };
 
 } // anonymous namespace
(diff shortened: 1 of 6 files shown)


src/config/calling_convention.cpp
```diff
@@ -166,5 +184,10 @@ std::ostream& operator<<(std::ostream &out, const CallingConvention& cc)
 	return out;
 }
 
+CallingConventionID CallingConvention::operator=(const CallingConventionID& cc)
```
s3rvac: C++ tip: Unless there is a specific reason for returning a value from an assignment operator, [it is generally recommended](http://www.icu-project.org/docs/papers/cpp_report/the_anatomy_of_the_assignment_operator.html) to return a reference:
> What does the function return? An assignment operator can return anything it wants, but the standard C and C++ assignment operators return a reference to the left-hand operand. This allows you to chain assignments together like so:
>
> ```cpp
> x = y = z = 3;
> ```
>
> Unless you have a really good reason, you want to follow this convention. Returning a reference to the right-hand operand or a value (i.e., another whole TFoo) would both still allow the simple chain described above to work, but have subtle differences in semantics from the way the standard operators do it that would come out in more complicated expressions involving assignment. Returning a value also forces unnecessary trips through the object's copy constructor, costing you in performance.
>
> So the outer shell of a properly-written assignment operator would look like this:
>
> ```cpp
> TFoo&
> TFoo::operator=(const TFoo& that)
> {
>     // copy the state...
>     return *this;
> }
> ```
>
> (By the way, note that the return statement is "return *this", rather than "return this". That's because this is a pointer to TFoo. You have to dereference it to get a reference to TFoo. Of course, if you forget this, the compiler will remind you.)

s3rvac: Also, an assignment operator for class `X` should generally return a reference to `X` and not some other type (e.g. a member as in your case).

xkubov: Thank you for the tip. I will incorporate the recommendation and update commit.

s3rvac:
One of the unit tests has failed on macOS in TravisCI ([link](https://retdec-tc.avast.com/viewLog.html?buildId=2686&buildTypeId=Retdec_MacBuild&tab=buildLog&_focus=4795)). Can you please take a look at it?

xkubov:
I looked at the failed unit test and the error message seems strange as the message comes from a part of RetDec that should not be affected by this PR at all. Moreover, I was unable to reproduce the same issue with unit tests on my macOS locally. I updated the branch by rebasing it over the latest master branch which triggered all tests to be run once more. If the issue will be present after all tests finish I will investigate it more.

s3rvac:
Thank you. I think this might be caused by nondeterminism in `capstone2llvmir` (or its tests). @PeterMatula can you please take a look?

PR #625, opened 2019-08-15
Improve the return value of utils::timestampToDate()

s3rvac:
When constructing the empty string, it is better to use `std::string()` (or just `{}` when returning a value) instead of `""` as the latter calls a construct that needs to traverse the empty string literal, which is needless pessimization.


s3rvac:
Run TeamCity builds.

PR #628, opened 2019-08-24
MIPS: Support orphaned lo16 relocs. Fixes #627

### LINKED ISSUE #627: MIPS: Support for relocations that pair multiple R_MIPS_LO16 against a single R_MIPS_HI16
According to https://www.linux-mips.org/pub/linux/mips/doc/ABI/mipsabi.pdf, a R_MIPS_LO16 without a R_MIPS_HI16 (orphaned) should be matched with the previous R_MIPS_HI16. This isn't currently taken into consideration in mips relocation.

This isn't a bug, but rather an enhancement. I think a simple variable to hold the previous value to pair with an orphaned lo16 would be sufficient here. If I get time I'll post a pull request for this.

bmourit:
MIPS: Add support for orphaned lo16 relocations.

This fixes issue #627

### DIFF
--- a/src/loader/loader/elf/elf_image.cpp
+++ b/src/loader/loader/elf/elf_image.cpp
@@ -761,6 +761,7 @@ void ElfImage::resolveRelocation(const retdec::fileformat::Relocation& rel, cons
 		case retdec::fileformat::Architecture::MIPS:
 		{
 			static const retdec::fileformat::Relocation* lastMipsHi16 = nullptr;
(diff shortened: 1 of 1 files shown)


PeterMatula:
Looks ok, lets Run TeamCity builds.

PeterMatula:
Well, some regression tests in TeamCity builds failed, but this is due to changes in regression tests suite and RetDec `master` branch - these repos changed, but changes were not merged to this PR. So this is probably nothing to worry about.

PR #670, opened 2019-10-18
cmake: added options to build only specific targets.

### LINKED ISSUE #510: Add option to install only specific CMake targets
RetDec is a large-scale project containing multiple standalone libraries and binaries. It might be useful if we were able to install only certain parts of RetDec, for example just `fileinfo` and its dependencies.

This might be achieved through `COMPONENT` attribute of `install` directives in CMake but I have never tried it myself. (https://cmake.org/cmake/help/v3.13/command/install.html)

PeterMatula:
Fix #510.


PeterMatula:
Run TeamCity builds.

PeterMatula:
Run TeamCity builds.

Issue #633, opened 2019-08-28
Unsuccessful MIPS decompilation


erorcun:
Hi, I've just discovered the tool and wanted to test it with some MIPS executable I have, but it seems decompilation being cut at some opcode (output is only ~5 lines) but the tool acts as it decompiled the whole function, without any error messages.

I've tried it within IDA and without it, but always with --select-decode-only and --select-range options. Also tried to shrink function end address gradually to see where's the problem, but I haven't had any luck.

I can send the binary file and range to any team member e-mail.

Thanks.

erorcun:
Why no team member replies to me?!

pfalcon:
> Why no team member replies to me?!

You may help the team by including number of your paid support contract, they must have forgotten it ;-)

xkubov:
Hi @erorcun. Sorry for the inconvenience. You can send the binary file privately to my email address (xkubov@gmail.com) and I (or someone from the team) will take a look at your issue.

xkubov:
Hi, I've received your file and looked into the issue. After I've run `retdec-decompiler.py` with the provided file, decompilation failed in debug mode on following assertion:
```
retdec-bin2llvmir: src/llvm-project/lib/IR/Instructions.cpp:2157: void llvm::BinaryOperator::AssertOK(): Assertion `getType()->isIntOrIntVectorTy() && "Tried to create a logical operation on a non-integral type!"' failed.
Error: Decompilation to LLVM IR failed
```

I've debugged the source of this issue and managed to find a bug in translating module `capstone2llvmir`. Specifically, a bug in the work with MSA (MIPS SIMD arch.) registers during LLVM IR creation.

**Little intro:**
[MIPS SIMD documentation](https://www.mips.com/products/architectures/ase/simd/) tells that size of registers `w0-w31` is 128 bits and each one of them can can be viewed (depending on the instruction) as:
 - 16 x 8bit elements
 - 8 x 16bit elements
 - 4 x 32bit elements
 - 2 x 64bit elements
Type of elements can be integer, fixed-point or floating-point.

In RetDec is each register implemented as 128-bit float and each instruction that works with such registers is required to extract required part of the register and cast it to the appropriate type (according to particular instruction) to create work with it.

**Back to the issue:**
[...]

PR #672, opened 2019-10-21
Try to fix issue #633

xkubov:
Redesigns `loadOp` methods that are responsible for the loading of operands of instructions.
Before this PR it was possible to load operands of incompatible type for the instruction which did
result in faulty decompilation. This PR tries to bring changes that will ensure that operands are
converted with a proper method before usage.

### DIFF
--- a/src/capstone2llvmir/capstone2llvmir_impl.h
+++ b/src/capstone2llvmir/capstone2llvmir_impl.h
@@ -185,7 +185,18 @@ class Capstone2LlvmIrTranslator_impl : virtual public Capstone2LlvmIrTranslator
 				llvm::Type* to,
 				eOpConv ct);
 
-		llvm::Type* checkTypeConversion(
+		/**
+		 * Internal method used to correct type used for operands
+		 * convertion based on specified "convertion type method" - ct.
+		 *
+		 * @param irb   LLVM IR Builder required for IR modifications.
+		 * @param to    result type that will be used to convert operands.
+		 * @param ct    convertion method by which will be opeands converted to the resut type.
+		 * @return      If result type for convertion can be used with specified conversion method
+		 *              returns param to. Otherwise will this method try to create suitable type
+		 *              for convertion method ct with size of llvm type of param to.
+		 */
+		llvm::Type* _checkTypeConversion(
 				llvm::IRBuilder<>& irb,
 				llvm::Type* to,
 				eOpConv ct);
--- a/src/capstone2llvmir/capstone2llvmir_impl.h
+++ b/src/capstone2llvmir/capstone2llvmir_impl.h
@@ -161,27 +161,32 @@ class Capstone2LlvmIrTranslator_impl : virtual public Capstone2LlvmIrTranslator
 			THROW,
 			/// Operand types does not have to be equal.
 			NOTHING,
-			/// Convert to destination type using ZEXT or TRUNC.
-			/// Types must be integer, or LLVM asserts.
-			ZEXT_TRUNC,
-			/// Convert to destination type using SEXT or TRUNC.
-			/// Types must be integer, or LLVM asserts.
-			SEXT_TRUNC,
-			/// Convert to destination type using FPCast (FPExt, BitCast,
+			/// Convert to destination integer type using ZEXT or TRUNC.
+			/// If source is FP type converts it using bitcast.
+			ZEXT_TRUNC_OR_BITCAST,
+			/// Convert to destination integer type using SEXT or TRUNC.
+			/// If source is FP type converts it using bitcast.
+			SEXT_TRUNC_OR_BITCAST,
+			/// Convert to destination FP type using FPCast (FPExt, BitCast,
 			/// or FPTrunc).
-			/// Types must be floating point, or LLVM asserts.
-			FP_CAST,
-			/// Convert to destination type using SIToFP.
+			/// If source is integer type converts it using bitcast.
+			FPCAST_OR_BITCAST,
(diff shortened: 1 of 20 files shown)


xkubov:
I guess let's run TeamCity.

PeterMatula:
Well, I guess you are not in that user list that can trigger the TC builds :-)

PeterMatula:
Run TeamCity builds.

PeterMatula:
Ok, @s3rvac says you are in the list, but the magic regex is `run (tc|teamcity) (builds|tests)`
ROFL

src/capstone2llvmir/capstone2llvmir_impl.h
```diff
@@ -325,24 +330,109 @@ class Capstone2LlvmIrTranslator_impl : virtual public Capstone2LlvmIrTranslator
...
+		std::vector<llvm::Value*> _loadOpsUniversal(
+				CInsn* ci,
+				llvm::IRBuilder<>& irb,
+				std::size_t opCnt,
+				bool strictCheck = true,
+				eOpConv ict = eOpConv::SEXT_TRUNC_OR_BITCAST,
+				eOpConv fct = eOpConv::FPCAST_OR_BITCAST);
 
-		llvm::Value* loadOpUnary(
+		llvm::Value* loadOpsUnary(
```
PeterMatula: Why `loadOpsUnary()` with `s`? There was `loadOpUnary()` before and there are `loadOpBinary()` now (with no `s`). Is this intentional? If so, what is the reason? Or is it a typo? If so, fix it back.

src/capstone2llvmir/capstone2llvmir_impl.h
```diff
@@ -161,29 +161,34 @@ class Capstone2LlvmIrTranslator_impl : virtual public Capstone2LlvmIrTranslator
...
+			UITOFP_OR_FPCAST
 		};
 
 		llvm::Value* generateTypeConversion(
 				llvm::IRBuilder<>& irb,
 				llvm::Value* from,
 				llvm::Type* to,
 				eOpConv ct);
+
+		llvm::Type* checkTypeConversion(
```
PeterMatula: What does this do? I see that it is used only in `_loadOps()`. There is no comment for the method. I don't like it much, because judging only by the name, it looks like something that should be called before/after(?) every type conversion (`generateTypeConversion()`). This would confuse me.

xkubov: This is meant to be only an internal method for the `_loadOps()` method. 

A request for this method originated in how `_loadOps()` is used. As the size of the type of loaded operands of translated instructions may vary from usage to usage, this method cannot be provided with specific `llvm::Type*` for loaded operands. Translation methods, however, know whether the full type will be integer type or floating-point type, and so the only information that they'll provide to the method `_loadOps()` is the conversion type. Thus it has to be able to deduce the type from the conversion method and the size of the type from the first loaded operand.

This deduction is provided by the `checkTypeConvertion` method that will be supplied with type and will return it as a prototype with a known size that can be used to convert other operands. If the type does not correspond to the used conversion method,  however, it tries to create type with equivalent size (either FP or integer - based on conversion type) and returns this type as a prototype for nest conversions.

Consider this example from issue 633:

The translation method of the instruction `ORI` receives two operands `A = Float128` and  `B = Float128`.
The translation method knows that its operands can only have integer type as it operates only on bits. This method does not care about the sign of integer operands. Conversion method that will ensure that every operand has equal size would be `ZEXT_OR_TRUNC`, and if we consider that by some mistake, a floating-point register may get into the input of this method, `ZERO_TRUNC_OR_BITCAST` will do the job. So this method uses `_loadOps` to load two integer operands of the `ORI` instruction. The method `_loadOps` now must deduce the size of operands, for example, by loading the first one.

Before this PR, the method has taken the first operand and declared its type as the result type. Other operands would get converted to this result type. As the type of first operands is Float128, this would result in incorrect type guessed as result type.

Now, the `_loadOps` method uses `checkTypeCovnertion` that will check whether it is possible to use the type of the first operand as example type (result type) or wheater it should be other types with the same size. So a combination of ZEXT_TRUNC_OR_BITCAST + Flaot128 operand will result in result type to be Int128, and both operands (A, B) will be converted to this type.

xkubov: I will provide documentation for this method and mark it as internal with an underscore (`_checkTypeConvertion`).

src/capstone2llvmir/capstone2llvmir_impl.h
xkubov: Oh, I forgot to change this one back. Originally I wanted to create new conversion methods, and this would help me identify all their usages. After a few methods I changed my mind, and I guess this was in another commit when I reverted the change. It is now a typo; I will change it back.

xkubov:
So I'll try to run TeamCity tests correctly now.

PeterMatula:
@xkubov, two more things:
1. Are there some tests for this? I know that it is a broad refactorization that aims to make the code that is already working more safe -- less error prone. But it was triggered by #633, so are there some test at least for the original issue reported there? If yes, then great. If no, add something please. Ideally, if possible, in a form of an unit test - it is the most straightforward and fastest running - you should be able to put in an exact instruction (in text or binary form) that caused the problem.
2. Add entry for the #633 fix, and this #672 PR to `CHANGELOG.md` - the same way as other entries there.

PeterMatula:
I wanted to finish this, so I did both of those things.

PR #683, opened 2019-11-28
Allow building against pog on 64-bit Linux machines

leftmostcat:
Current master will fail to build on 64-bit Linux machines due to looking for `libpog_fmt.a` at `external/src/yaramod-project-build/deps/pog/pog-install/lib/libpog_fmt.a` whereas, due to `include(GNUInstallDirs)` in the pog cmake config, the file will be located at `external/src/yaramod-project-build/deps/pog/pog-install/lib64/libpog_fmt.a`. This can be solved by adding the `include` statement (already present in `yaramod`) and using `CMAKE_INSTALL_LIBDIR` in the pog path.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,9 @@
 cmake_minimum_required(VERSION 3.6)
 
 project(retdec C CXX)
+
+include(GNUInstallDirs)
+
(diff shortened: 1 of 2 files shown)


metthal:
I've merged it. Thank you for your contribution.

PR #699, opened 2020-02-04
Request to provide new function type

xkubov:
Hi in this PR I want to request and discuss a design change of function type in RetDec config.

Before this PR function type might have been one of the following:
  - USER_DEFINED
  - STATICALLY_LINKED
  - DYNAMICALLY_LINKED
  - SYSCALL
  - IDIOM

The default function type that was generated for every function was
USER_DEFINED. This in my opinion is not right as we have to distinguish functions that
were provided by user through some kind of front-end and functions that
were automatically detected by decompiler. This is why I propose new type
- DECOMPILER_DEFINED.

A reason to distinguish these two function types is for example in parameter
analysis. When user specifies that the return type is integer, the decompiler
should not continue to do heuristics to detect return type. If an user
wants a function to be of some type they should get the function with the type
on output. Also this brings more flexibility when it comes to creation of custom plugins
that provide data to RetDec (like r2plugin).

I traced every implementation/usage of previous default type (USER_DEFINED) and changed it accordingly.

### DIFF
--- a/include/retdec/common/function.h
+++ b/include/retdec/common/function.h
@@ -35,9 +35,23 @@ using LineNumber = retdec::common::Address;
 class Function : public retdec::common::AddressRange
 {
 	public:
+		/**
+		 * Recognized types of a function that will determine
+		 * how the decompiler will treat the specified function.
+		 *
+		 * When the type is DECOMPILER_DEFINED the decompiler is
+		 * allowed to prefer info recieved from some heuristics,
+		 * instead of info specified in the config.
+		 *
+		 * When the type is USER_DEFINED the info about the function
+		 * (params, type) specified in a config file will be projected
+		 * on the decompiler output and the decompiler should not do
+		 * any heuristcs.
+		 */
 		enum eLinkType
 		{
-			USER_DEFINED = 0,
+			DECOMPILER_DEFINED = 0,
+			USER_DEFINED,
 			STATICALLY_LINKED,
 			DYNAMICALLY_LINKED,
 			SYSCALL,
@@ -53,6 +67,7 @@ class Function : public retdec::common::AddressRange
 
 		/// @name Function query methods.
 		/// @{
+		bool isDecompilerDefined() const;
 		bool isUserDefined() const;
 		bool isStaticallyLinked() const;
 		bool isDynamicallyLinked() const;
@@ -80,6 +95,7 @@ class Function : public retdec::common::AddressRange
 		void setWrappedFunctionName(const std::string& n);
 		void setStartLine(const retdec::common::Address& l);
 		void setEndLine(const retdec::common::Address& l);
+		void setIsDecompilerDefined();
 		void setIsUserDefined();
 		void setIsStaticallyLinked() const;
 		void setIsDynamicallyLinked() const;
@@ -134,7 +150,7 @@ class Function : public retdec::common::AddressRange
 		std::string _declarationString;
 		std::string _sourceFileName;
 		std::string _wrapperdFunctionName;
-		mutable eLinkType _linkType = USER_DEFINED;
+		mutable eLinkType _linkType = DECOMPILER_DEFINED;
 		LineNumber _startLine;
 		LineNumber _endLine;
 		bool _fromDebug = false;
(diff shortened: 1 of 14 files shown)


xkubov:
Meanwhile, lets Run TeamCity builds and see if everything is ok.

xkubov:
Well, I will look into the problem why tests in debug build did not pass and push fix. However the doxygen build is failing becuase of the new integrated libraries (yaracpp, pelib). I can fix those too, but I think we should discuss if we will not disable building documentation for those (for example by moving these libraries back into deps/ directory).

xkubov:
Lets Run TeamCity builds after doxygen build and tests on debug build on master are passing.

include/retdec/common/function.h
```diff
@@ -53,6 +54,7 @@ class Function : public retdec::common::AddressRange
 
 		/// @name Function query methods.
 		/// @{
+		bool isDecompilerDefined() const;
 		bool isUserDefined() const;
```
s3rvac: After this PR is merged, could you please prepare a PR for our [IDA plugin](https://github.com/avast/retdec-idaplugin)? It relies on `isUserDefined()` (see [this code](https://github.com/avast/retdec-idaplugin/blob/1b712167bbad1b1d30671a93af74fd3ae5a68d63/src/idaplugin/code_viewer.cpp#L296)).

include/retdec/common/function.h
```diff
@@ -37,7 +37,8 @@ class Function : public retdec::common::AddressRange
 	public:
 		enum eLinkType
 		{
-			USER_DEFINED = 0,
+			DECOMPILER_DEFINED,
+			USER_DEFINED,
```
s3rvac: Consider adding a comment describing the difference between `DECOMPILER_DEFINED` and `USER_DEFINED`. We know the difference now, but in the future, we or anybody else may wonder about that.

include/retdec/common/function.h
```diff
@@ -53,6 +54,7 @@ class Function : public retdec::common::AddressRange
```
s3rvac: _(A question not related to the attached line.)_

> When user specifies that the return type is integer, the decompiler should not continue to do heuristics to detect return type.

This will be implemented in a separate PR, right?

src/bin2llvmir/optimizations/dsm_generator/dsm_generator.cpp
```diff
@@ -225,7 +225,7 @@ void DsmGenerator::generateFunction(
 			<< fnc->getStart().toHexPrefixString()
 			<< " -- " << fnc->getEnd().toHexPrefixString() << "\n";
 
-	if (!fnc->isUserDefined())
+	if (!(fnc->isDecompilerDefined() || fnc->isUserDefined()))
```
s3rvac: Consider moving the negation into the or-expression to make the code a bit more readable:
```suggestion
	if (!fnc->isDecompilerDefined() && !fnc->isUserDefined())
```

src/llvmir2hll/ir/module.cpp
```diff
@@ -374,6 +374,30 @@ bool Module::hasFuncDefinitions() const {
...
+*/
+FuncSet Module::getDecompilerDefinedFuncs() const {
+	return getFuncsSatisfyingPredicate(
+		[this](auto func) {
+			return config->isDecompilerDefinedFunc(func->getInitialName());
+		}
+	);
+}
+
+
```
s3rvac: Nitpick: Too many newlines (one is enough).

include/retdec/common/function.h
xkubov: Yes of course, thank you for the notice. Updating plugins is on my list after this. Apart from this IDA plugin should get more updates as @PeterMatula introduced changes to RetDec config.

include/retdec/common/function.h
xkubov: Yes, you are right. I think it would be good idea, too.

include/retdec/common/function.h
xkubov: Yes it will be. I have this implemented on separate branch `config-user-info`. I wanted this PR to provide a new function type only.

src/bin2llvmir/optimizations/dsm_generator/dsm_generator.cpp
xkubov: Okay.

src/llvmir2hll/ir/module.cpp
xkubov: Hm, I would swear I removed it from the commit. Nevertheless, thank you I'll fix this.

PR #715, opened 2020-02-19
Extension of bin2llvmir with optimization of X87 FPU stack.

JurajHolub:
Reimplementation of current FPU stack analysis. The new design supports dependencies of CFG for functions and BasicBlocks. Optimization uses 3rd party library Eigen for calculation with matrices. The new component also includes a set of unit tests.


deps/eigen/README.md
```diff
@@ -0,0 +1,8 @@
+## Eigen
+
+A clone of only the needed sources (headers) of [Eigen](http://eigen.tuxfamily.org/index.php?title=Main_Page).
```
s3rvac: Just so we don't forget: We should discuss with @PeterMatula whether it would be possible to use an external project and download the sources of Eigen when building RetDec so we don't need to have all the 3rd-party sources in our repo. We do this for many of our dependencies.

xkubov:
Are you sure you pushed the correct version?

As you can see there is a build error in CI:

https://travis-ci.org/avast/retdec/jobs/652520785?utm_medium=notification&utm_source=github_status

xkubov:
Also when fixed the redefinition error:

/home/kubov/projects/retdec/src/bin2llvmir/optimizations/x87_fpu/x87_fpu.cpp:19:10: fatal error: libalglib/ap.h: No such file or directory
 #include <libalglib/ap.h>
          ^~~~~~~~~~~~~~~~
compilation terminated.

Please fix this and update the PR.

Also, may I ask why should be Eigen included in the sources? Can't we demand it to be installed as a RetDec dependency? Do we need a specific or patched version of Eigen?

s3rvac:
> Also, may I ask why should be Eigen included in the sources? Can't we demand it to be installed as a RetDec dependency? Do we need a specific or patched version of Eigen?

For the record, this is already being discussed [here](https://github.com/avast/retdec/pull/715#discussion_r381358296).

JurajHolub:
Actual state fixes previous bugs but the Eigen library is still included in the sources. As soon as possible, I change it to project dependency to install.

PeterMatula:
Run TeamCity builds.

PeterMatula:
run tc tests

PR #719, opened 2020-03-06
Lz signature zeroed check

ladislav-zezula:


s3rvac:
Let's run TeamCity builds to see whether everything succeeds in there.

src/pelib/PeLibAux.cpp
```diff
@@ -135,6 +139,23 @@ namespace PeLib
 		return (ByteSize >> PELIB_PAGE_SIZE_SHIFT) + ((ByteSize & (PELIB_PAGE_SIZE - 1)) != 0);
 	}
 
+	bool pelibIsBufferZeroed(const void * buffer, size_t length)
```
s3rvac: Is it necessary to prefix the member function with "pelib"? Couldn't we name it just `isBufferZeroed()`?

src/pelib/PeLibAux.cpp
```diff
@@ -135,6 +139,23 @@ namespace PeLib
...
+	{
+		const uint32_t * buffer32 = reinterpret_cast<const uint32_t *>(buffer);
```
s3rvac: The implementation seems unnecessarily complicated and I had a hard time understanding it. Couldn't we replace it with this tiny template (in `include/retdec/pelib/PeLibAux.h`)?
```cpp
template<typename T>
bool pelibIsBufferZeroed(const std::vector<T>& buffer)
{
    return std::all_of(buffer.begin(), buffer.end(), [](T item) { return item == 0; });
}
```
The prototype of the function is simpler, we no longer need to separately pass the buffer and its size, there are no reinterpret casts, and the code is much more readable (it just checks that all items in the buffer are zero, just like you would expect).

Or does the implementation do something other than checking if the array contains only zeros?

Issue #702, opened 2020-02-08
Regression tests on macOS sometimes randomly fail


s3rvac:
[This macOS build](https://retdec-tc.avast.com/viewLog.html?buildId=3418&buildTypeId=Retdec_MacBuild) has succeeded. The one [after that](https://retdec-tc.avast.com/viewLog.html?buildId=3425&buildTypeId=Retdec_MacBuild), which is for a commit that only changed the changelog file, failed because of the following test:
```
bugs.selective-decompilation-strange-behaviour.TestDecompileSelect0x0DecodeAll (main.ex --sel[..])  [ FAIL ]  (1.15s)
F
======================================================================
FAIL: test_has_no_functions (bugs.selective-decompilation-strange-behaviour.TestDecompileSelect0x0DecodeAll)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "retdec-regression-tests-framework/regression_tests/tools/decompiler_test.py", line 65, in setUp
    self.assertEqual(self.decompiler.return_code, 0, msg=msg)
AssertionError: 1 != 0 : decompiler failed; output:
...
Running phase: Simple types recovery optimization ( 0.65s )
Running phase: Stack pointer operations optimization ( 0.65s )
Running phase: Instruction idioms optimization ( 0.65s )
Running phase: LLVM ( 0.65s )
Running phase: LLVM instruction optimization ( 0.65s )
Running phase: Instruction idioms optimization ( 0.65s )
Running phase: Phi removal ( 0.65s )
Running phase: Value protection optimization ( 0.65s )
Error: Decompilation to LLVM IR failed
```
I [restarted the build](https://retdec-tc.avast.com/viewLog.html?buildId=3430&buildTypeId=Retdec_MacBuild&tab=buildLog) after that, and it failed again, but because of another test:
```
integration.palindrome.Test_2017 (palindrome.x86.clang.O0.g.elf)                                    [ FAIL ]  (0.85s)
F
======================================================================
FAIL: test_produce_expected_output (integration.palindrome.Test_2017)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "retdec-regression-tests/integration/palindrome/test.py", line 13, in test_produce_expected_output
    expected_output='krk is a palindrome\n'
  File "retdec-regression-tests-framework/regression_tests/tools/decompiler_test.py", line 97, in assert_c_produces_output_when_run
    self.assertEqual(actual_output, expected_output, msg)
[...]

PR #734, opened 2020-04-08
Add detection support for newer versions of VMProtect.

tamaroth:
Since version 2.04, every protected file has a public signature (detected with a YARA rule) and
the checksum of the first 64 DWORDS in the last executable section is always constant despite the
bytes in that part being different.

In addition, the tests were added [here](https://github.com/avast/retdec-regression-tests/pull/46)


s3rvac:
Let's run TeamCity tests.

src/cpdetect/compiler_detector/heuristics/pe_heuristics.cpp
```diff
@@ -1142,6 +1142,8 @@ void PeHeuristics::getExcelsiorHeuristics()
  */
```
s3rvac, line 1142: _(A comment not related to the attached line.)_

The following regression tests have failed [in this TC build](https://retdec-tc.avast.com/viewLog.html?buildId=3716&buildTypeId=Retdec_LinuxBuild&tab=buildLog):
```
tools.fileinfo.features.section-name-packer-compiler-detection.TestRamnit (b9bf69972077eee9c6[..])  [ FAIL ]  (0.14s)
tools.fileinfo.features.vmprotect-heuristic.Test (TidyESP.ex)                                       [ FAIL ]  (0.22s)
```
Could you please take a look at them? The failures seem to be caused by the changes in the present PR.

s3rvac: You can run them in the following way:
```
$ ./runner.py tools.fileinfo.features.section-name-packer-compiler-detection -r TestRamnit
$ ./runner.py tools.fileinfo.features.vmprotect-heuristic -r TidyESP
```

s3rvac:
Let's run TeamCity tests.

src/cpdetect/compiler_detector/heuristics/pe_heuristics.cpp
s3rvac: The second test has been fixed. As for the first test, it fails only in TeamCity and only in a debug Linux build. For some reason, `fileinfo.return_code` is 245. I have no idea why. I will have to dig deeper.

tamaroth: It was probably caused by the following code:

```
auto data = section->getBytes().data();
```
The `.data()` returns a reference to already existing data, but that data gets destroyed after that code finished executing and the reference is invalidated.

To fix this I have moved the result of the `section->getBytes()` to a separate variable, which will only be destroyed when its scope ends.

s3rvac:
Let's run TeamCity tests.

s3rvac:
That [macOS fail](https://retdec-tc.avast.com/viewLog.html?buildId=3738&buildTypeId=Retdec_MacBuild&tab=buildLog&_focus=9462) is caused by #702 (not by the present PR). However, the [Linux build](https://retdec-tc.avast.com/viewLog.html?buildId=3737&buildTypeId=Retdec_LinuxBuild&tab=buildLog&_focus=10689) is unfortunately still failing. I will try to debug it. There is either another bug, or the build did not run with the fix, or there is something else going on.

s3rvac:
The build has run for ec1cc47c24003329373e540dd72393a14e7ccc24, which is currently the latest commit in this PR. So, the build has run with the fix.

s3rvac:
I am able to reproduce the issue when using [this Dockerfile](https://github.com/avast/retdec/blob/master/Dockerfile). However, I was not yet able to determine the exact cause as when I start debugging the issue, the [problems seem to go away](https://en.wikipedia.org/wiki/Heisenbug). I will investigate it further. I just wanted to let you know about the progress.

s3rvac:
Let's run TeamCity tests.

s3rvac:
Alright, so, even after about 6 hours of debugging, I was unable to determine the real cause. Here is a couple of notes:

* `retdec-fileinfo` was crashing due to the changes in `PeHeuristics::getVmProtectHeuristics()` in this PR. Here is a backtrace from `gdb`:
    ```
    #0  0x00005555558618b8 in retdec::cpdetect::PeHeuristics::getVmProtectHeuristics() ()
    #1  0x000055555586f531 in retdec::cpdetect::PeHeuristics::getFormatSpecificCompilerHeuristics() ()
    #2  0x000055555583ec12 in retdec::cpdetect::CompilerDetector::getAllCompilers() ()
    #3  0x000055555584050b in retdec::cpdetect::CompilerDetector::getAllInformation() ()
    #4  0x0000555555666ee1 in retdec::fileinfo::FileDetector::getCompilerInformation() ()
    #5  0x00005555556677ac in retdec::fileinfo::FileDetector::getAllInformation() ()
    #6  0x0000555555614bc0 in main ()
    ```
    Line information is not available as the crashes are only reproducible in a release build.
* The crash was only reproducible with GCC 7 in Ubuntu (this is what the Linux TC agent runs). I was able to reproduce it by using [this Dockerfile](https://github.com/avast/retdec/blob/master/Dockerfile).
* The crash was only reproducible with the `Release` build type. If I tried a different build type (even `RelWithDebInfo`), it stopped crashing.
* I have tried upgrading GCC 7 from 7.4.0 to 7.5.0, but to no avail (it did not fixed the crashing).
* `valgrind` did not report any errors apart from `retdec-fileinfo` crashing with the backtrace above. I have tried several different build types with several different compilers.
* When I tried to debug the code by manually inserting debug prints, the crashing stopped. The crashing also stopped when I modified the computing of the checksum.
* The crash appeared to happen on the following line:
    ```cpp
    checksum += *reinterpret_cast<const std::uint32_t*>(&data[i * 4]);
    ```
  However, when I simply printed the obtained integers, the code worked without any issues.
* The memory returned by `section->getBytes()` and stored in `data` seemed to be valid.

I do not know how to debug this further. I believe this to be a compiler bug. I would really like to know this for sure, but I do not have the time nor will to debug it further.

Nevertheless, I was able to modify the code so it is more readable and the Linux TC build no longer fails. @tamaroth: If the current code is OK from your point of view, we can merge the PR as all the tests pass now.

src/cpdetect/compiler_detector/heuristics/pe_heuristics.cpp
s3rvac: This did not fix the crashing. For more details, see [this comment](https://github.com/avast/retdec/pull/734#issuecomment-611977060). All the tests pass now, so I am marking this as resolved.

tamaroth:
Thank you for your thorough investigation. If it crashed only on a specific version and only when placed directly in the checksum computing loop, perhaps there was a bug in optimisation? Hard to tell.

Again, thank you for your code review and updated code, it will be useful in the future.

The code looks good to me! :+1:

PR #735, opened 2020-04-09
Added tests for more versions of Petite packer

ladislav-zezula:


s3rvac:
Let's run TeamCity tests.

PR #741, opened 2020-04-15
Improve the detection of Enigma protector (32-bit).

tamaroth:
The following changes ensure that previously undetected versions, and
hopefully future versions as well, are properly detected as Enigma.

See avast/retdec-regression-tests/pull/48 for tests.


s3rvac:
Let's run TeamCity tests.

PR #743, opened 2020-04-16
Improve the detection of ASPack executable packer.

tamaroth:
The following changes ensure that previously undetected versions of ASPack are properly detected.

Tests added here: avast/retdec-regression-tests/pull/49


s3rvac:
Let's run TeamCity tests.

PR #745, opened 2020-04-20
Fixed missing <limits> header

Cyclic3:
On my machine (Debian/sid with g++ 9.3.0 and glibcxx 3.4.28), the header <limits> does not automatically get included, leading to a large number of compilation errors. 

This fixes that problem (at least on my machine).


s3rvac:
Thank you! Let's run TeamCity tests to see if everything passes.

PR #746, opened 2020-04-20
Improve the detection of the Eziriz packer/protector

ladislav-zezula:


support/yara_patterns/tools/pe/x86/packers.yara
```diff
@@ -70,6 +70,72 @@ rule ep_exepack_360_406 {
...
+		$s02 = "{11111-22222-30001-00001}" wide
+		$s03 = "{11111-22222-40001-00001}" wide
+		$s04 = "{11111-22222-50001-00001}" wide
+	condition:
+		pe.number_of_sections == 3 and
+		pe.imports("mscoree.dll") and
+		all of them
+}
+
+// TODO: When retdec's YARAC will be of a newer version
```
s3rvac, line 122: What version do we need? We use YARA 3.11.0, which is currently (2020-04-20) the latest stable release. Maybe you have discussed this with other people from the team, but I would also like to know :slightly_smiling_face:

s3rvac: Oh, I see it is because of #747.

s3rvac:
Thank you for the fixes. Let's run TeamCity tests.

PR #748, opened 2020-04-21
Added detection of PyInstaller

ladislav-zezula:


s3rvac:
Let's run TeamCity tests.

src/cpdetect/compiler_detector/compiler_detector.cpp
```diff
@@ -434,7 +434,7 @@ ReturnCode CompilerDetector::getAllSignatures()
 					{
 						toolInfo.addTool(
 							DetectionMethod::YARA_RULE,
-							metaToStrength(rule.getMeta("language")),
+							metaToStrength(rule.getMeta("strength")),
```
s3rvac, line 437: This is probably just a bug fix of code from #746 and is not related to the detection of PyInstaller, right?

ladislav-zezula: Yes.

PR #753, opened 2020-04-24
Add detection support of Astrum InstallWizard.

tamaroth:
Astrum InstallWizard is an installation software from Thraex Software and this PR aims at providing Fileinfo an ability to detect installation files, created using that software, correctly.

The tests for this enhancment were added [here](/avast/retdec-regression-tests/pull/53).


s3rvac:
Let's run TeamCity tests.

PR #756, opened 2020-04-29
Add detection of AutoHotKey compiler.

tamaroth:
AutoHotKey is a script compiler similar to AutoIt that aims at
automating and imitating the behaviour of an end user via series
of macros, hotkeys, and software automations.


s3rvac:
Let's run TeamCity tests.

PR #761, opened 2020-05-04
The detection of BAT to PE-EXE script-compilers.

tamaroth:
This pull request adds the detection of the following BAT to PE-EXE compilers:

* [F2KO BAT2EXE](https://www.f2ko.de/programme/bat-to-exe-converter/)
* [Advanced BAT to EXE Converter](https://www.battoexeconverter.com/buypro.html)
* ExeScript (discontinued... website not available anymore)

These programs are not strictly compilers, they are quite difficult to categorise. They never compile the BAT to a VM or intermediate language, but rather the store the script file in the executable file (resources, read-only data, overlay, etc.) and execute them (by either dropping them to a file or seldomly directly from memory).

If required, I could change their categorization from `Compiler` to `Packer`, despite the fact they rarely use compression on the batch files.

The corresponding tests have been added to `retdec-regression-tests`.


s3rvac:
Let's run TeamCity tests.

PR #763, opened 2020-05-04
deps/yara: force rebuild if config changed, fix #760

### LINKED ISSUE #760: Compilation of YARA rules during an incremental build fails with "error: unknown module dotnet"
Compilation of YARA rules during an incremental build fails with `error: unknown module "dotnet"`.

## Steps to reproduce

For Linux, but the bug is platform-independent:
```
$ git clone https://github.com/avast/retdec
$ cd retdec
$ git checkout 8bcd65017faf195467a219f33ebc581623ae4652 # A commit prior to adding support for the dotnet module (#749).
$ mkdir build && cd build
$ cmake .. -DCMAKE_INSTALL_PREFIX=install
$ make -j$(nproc) && make install -j$(nproc)
$ git checkout master # Or any commit after adding support for the dotnet module (#749).
$ make -j$(nproc) && make install -j$(nproc)
```

## Expected output

The incremental build and installation succeed.

## Actual output

```
-- Up-to-date: yarac version '3.11.0'
-- Up-to-date: retdec/build/install/share/retdec/support (version is OK)
-- yarac version '3.11.0' written to 'retdec/build/install/share/retdec/support/version-yarac.txt'
-- Installing: retdec/build/install/share/retdec/support/generic/yara_patterns/tools/pe/x86/packers.yara
-- Compiling: retdec/build/install/share/retdec/support/generic/yara_patterns/tools/pe/x86/packers.yara
retdec/build/install/share/retdec/support/generic/yara_patterns/tools/pe/x86/packers.yara(7): error: unknown module "dotnet"
retdec/build/install/share/retdec/support/generic/yara_patterns/tools/pe/x86/packers.yara(115): error: invalid field name "number_of_user_strings"
Error: yarac failed during compilation of file retdec/build/install/share/retdec/support/generic/yara_patterns/tools/pe/x86/packers.yara
```

## Probable cause and notes

Commit dfc58a69a1c6fe2aa67d398179dad6be7b6ec042 (#749) added support for building YARA with the `dotnet` module. However, YARA (`yarac`) version was left unchanged. This made the build system believe that `yarac` does not need to be re-built when doing an incremental build. However, without re-building, the `dotnet` module is unavailable.

A workaround is to remove `build/deps/yara`, but this is annoying. Also, when someone stumbles upon this bug, they may wonder whether they are doing something wrong, or whether there is something wrong with RetDec. Last, but certainly not least, this bug will probably bite us again in the future.

My suggestion is to improve the build system so it detects changes in the build parameters and re-builds `yarac` when we change the build parameters. We could bump the version of YARA, but this feels more like a hack (we have not changed anything in YARA, except for adding a new build parameter).

Tested on both 64-bit Arch Linux and Debian Stretch with GCC 9.2.0. However, the bug is platform-independent.

PeterMatula:


PeterMatula:
lets run tc tests

PeterMatula:
@s3rvac can you have a look? It works for me, but this kind of things are tricky, so I would feel better if someone else tried it and looked at the solution.

Add solution: it is complicated by:
* external project configuration is different for MSVC than for other environments
* YARA is not build using CMake, but in source using make/MSVC solution
* simply triggering rebuild without cleaning does not work. CMake will not download the package again [and clean everything] because it did not change, and in-source build will not properly rebuild if it was already built.

s3rvac:
Sure, I will take a look.

PR #767, opened 2020-05-15
Lz more corruptions

ladislav-zezula:


PeterMatula:
@ladislav-zezula the code look ok.

From the user point of view, do I understand it correctly that this expands error codes returned by `FileFormat::getLoaderErrorInfo()`? Or does this add some separate error code that need to be get in some other way?
I.e. do I need to add a new entry to tools that use this kind of info, or is already ok if there is an entry for `getLoaderErrorInfo()`?

PR #773, opened 2020-05-23
deps/yara/patch: patch YARA surces to fix bug in v4.0.1

PeterMatula:
Bug being fixed: https://github.com/VirusTotal/yara/pull/1289/files
This needs to be removed once the fix gets to RetDec.


PeterMatula:
run tc tests

PR #778, opened 2020-06-02
Remove too broad YARA rules for VMProtect packer detection.

tamaroth:
These patterns are very common in regular files and do not indicate the presence of
VMProtect packer. In the `cpdetect` module there is a heuristic check that verifies the
same code alongside additional checks to ensure the detected packer is correct.


s3rvac:
Let's run TC tests.

PR #781, opened 2020-06-04
Improve the detection of CExe packer by modifying the YARA rule.

tamaroth:
The new YARA rule is more inclusive and correct one bug by changing the
variable byte in the call byte.

The change is twofold, the call 

```
FF 15 58 10 40 00                             call    ds:GetModuleFileNameA
```
had its most significant byte set to a wildcard, while in reality, it should have been the least significant byte instead.

```suggestion
<<< FF 15 58 10 40 ??
>>> FF 15 ?? 10 40 00
```

Finally, there is a version that has the final jump of the detecting code jump slightly closer.

```
74 16
74 14
```

Since it's the last byte, it's worth simply removing it.

The change is minute, not affecting current detections, therefore I have not added any tests.

If you think tests should be added, please let me know.


s3rvac:
Let's run TC tests.

PR #782, opened 2020-06-04
Very basic processing of UTF-16 to UTF-8 in PE resource table

metthal:
Do not merge for now, just trying to introduce very basic UTF-16 to UTF-8 transcription.

### DIFF
--- a/src/utils/string.cpp
+++ b/src/utils/string.cpp
@@ -6,13 +6,16 @@
 
 #include <algorithm>
 #include <cassert>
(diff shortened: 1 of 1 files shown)


metthal:
I suggest leaving it as it is right now/closing it. The implementation itself is not that hard, I think the code in this PR actually worked like it should. What's more problematic are the implications of this change on all the users of `retdec-fileinfo` and that would take 10x more effort than this alone.

PR #783, opened 2020-06-10
fileformat/format_detection: use file magic to detect COFF files.

PeterMatula:
Do not parse COFF files at format detection stage - this could be slow if the file is big.


PeterMatula:
run tc tests

PeterMatula:
run tc tests

PR #804, opened 2020-07-13
Improvement to the detection of multiple packers.

tamaroth:
This PR changes the following:

**Added**:
* Added new version of `Ste@lth` packer.

**Removed**:
* Ramnit was removed as a packer (it is a file-infector).
* Several ambiguous YARA rules that were too generic and not specific to described packer (`Themida`, `InstallShield`, `MEW`, `tElock`, `Ste@lth`).
* Redundant `AHPack` rule since `AHPack` and `!EP ExeCrypt` are the same product with different names.
* Incorrect recognition of `RLPack`. It was detecting code that was generated by `Themida`

**Improved**:
* Detecting UPX by section names now only detects when the very first section is zero-size.
* UPX detection based on the magic block now only detects when this block is placed before the first section.
* The detection of VMProtect is now more precise, lowering the number of false-positive detections.
* Renamed `WinUpack` to `Upack` for consistency.
* Renamed `yoda's Crytper` to `yoda's Protector` for consistency.
* Renamed one `ASPack` rule to `ASProtect` as it was detecting the code that is only present in `ASProtect`.

[Corresponding tests](https://github.com/avast/retdec-regression-tests/pull/65)


s3rvac:
Let's run TC tests.

PR #807, opened 2020-07-14
deps/openssl: remove, require it as a prerequisite.

PeterMatula:
Also refactor (i.e. remove) RetDec's Crypto library.


PeterMatula:
lets run TC tests

PR #816, opened 2020-07-22
Provide unified logging interface

xkubov:
In this PR I would like to introduce changes to the RetDec decompiler logging system. Before these changes each module controlled output on stdout/stderr on its own.

High-level changes:

* Added option `--silent` to turn of logging on stdout.
* Added option `"logFile"` to the retdec config.
* Added option `"errFile"` to the retdec config.

Changes above are particularly useful for users of the RetDec library. Before this PR there was no easy way to redirect RetDec's output when decompilation was called from `retdec::decompiler` function.

Logging design:
* Created interface `retdec::utils::io::Log` with functions `info()`, `error()`, `set(...)` that control logging objects. More info in documentation.
  - Usage:
    - `Log::info() << Log::Color::Yellow << "RetDec module" << std::endl`
    - `Log::error() << Log::Warning << "Unable to allocate stuff" << std::endl` 

With these changes, colors on output are managed in RetDec (before we used LLVM library). Logger class implemented to print output uses ANSI encoding for colors. This is supported on all POSIX terminals and Windows terminals following the Anniversary update. Changes were tested on Linux, macOS, and Windows. Colors are automatically disabled when output is not a terminal.


xkubov:
Let's run TeamCity builds

xkubov:
Fixed error, let's run TeamCity builds again.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 # dev
 
+* Enhancement: Unified logging on stdout/stderr. Added option `--silent`. Printed text is colored only when output is a terminal ([#791](https://github.com/avast/retdec/issues/791).
```
PeterMatula, line 5: As @s3rvac suggested in some other place, it is better to add these entries after the merge to `master`. It is then possible to reference the PR.

s3rvac: + the risk of running into conflicts is much lower :slightly_smiling_face:

PR #817, opened 2020-07-22
* Fixed high memory usage on samples with screwed up delayed imports

ladislav-zezula:
* ImageLoader::stringLength() now stops properly on zero pages

### DIFF
--- a/include/retdec/pelib/DelayImportDirectory.h
+++ b/include/retdec/pelib/DelayImportDirectory.h
@@ -103,7 +103,7 @@ namespace PeLib
 				init();
 
 				// Keep loading until we encounter an entry filles with zeros
-				for(std::size_t i = 0;; i += sizeof(PELIB_IMAGE_DELAY_LOAD_DESCRIPTOR))
+				for(std::uint32_t i = 0;; i += sizeof(PELIB_IMAGE_DELAY_LOAD_DESCRIPTOR))
 				{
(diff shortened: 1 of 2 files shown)


include/retdec/pelib/DelayImportDirectory.h
```diff
@@ -113,6 +113,10 @@ namespace PeLib
 					if(!imageLoader.readImage(&rec.delayedImport, rva + i, sizeof(PELIB_IMAGE_DELAY_LOAD_DESCRIPTOR)))
 						break;
 
+					// Valid delayed import entry starts either with 0 or 0x01
+					if(rec.delayedImport.Attributes & 0xFFFF0000)
+						break;
```
PeterMatula, line 118: From the changes in this PR, this one seems the only one that adds a "sanity break" defending against invalid import entries.
1. Am I right?
2. If so, can you compare it against the original solution? I.e. before #812 was merged? Is it the same?
I guess the main question is, does this solution solve only the problems in samples found by @s3rvac, or is it the same as the original approach and should behave the same?

ladislav-zezula: 1. Not exactly. There was also quite serious bug in ImageLoader::stringLength(), which also fixes loading strings from zeroes pages.

By all means I'll check against the pre-merge version.

s3rvac: I have tried running the fixed version on the found samples and it has fixed the high-memory issues.

PR #819, opened 2020-07-23
fileformat: Include pkcs7.h instead of doing forward declarations.

jobermayr:
Fixes:
In file included from ../retdec/src/fileformat/file_format/pe/pe_format.cpp:24:
../retdec/include/retdec/fileformat/file_format/pe/pe_format.h:23:26: error: conflicting declaration ‘typedef struct evp_md_st EVP_MD’
   23 | typedef struct evp_md_st EVP_MD;
      |                          ^~~~~~
In file included from /usr/include/openssl/crypto.h:129,
                 from /usr/include/openssl/bio.h:69,
                 from /usr/include/openssl/asn1.h:67,
                 from ../retdec/src/fileformat/file_format/pe/pe_format.cpp:15:
/usr/include/openssl/ossl_typ.h:113:26: note: previous declaration as ‘typedef struct env_md_st EVP_MD’
  113 | typedef struct env_md_st EVP_MD;
      |                          ^~~~~~

### DIFF
--- a/include/retdec/fileformat/file_format/pe/pe_format.h
+++ b/include/retdec/fileformat/file_format/pe/pe_format.h
@@ -17,10 +17,7 @@
 #include "retdec/fileformat/types/dotnet_types/dotnet_class.h"
 #include "retdec/fileformat/types/visual_basic/visual_basic_info.h"
 #include "retdec/pelib/PeFile.h"
-
-// Forward declare OpenSSL structures used in this header.
-typedef struct pkcs7_st PKCS7;
-typedef struct evp_md_st EVP_MD;
+#include "openssl/pkcs7.h"
 
 namespace retdec {
 namespace fileformat {


PeterMatula:
There is a lot of issues, so sorry if I missed something, but is there an issue regarding this compilation error?
```
In file included from ../retdec/src/fileformat/file_format/pe/pe_format.cpp:24:
../retdec/include/retdec/fileformat/file_format/pe/pe_format.h:23:26: error: conflicting declaration ‘typedef struct evp_md_st EVP_MD’
```
Can we try to investigate why it is happening and possibly fix it in some other way? Either on your side, or in our sources (e.g. not using these OpenSSL structures in headers at all).

The thing is, that we are intentionally not including OpenSSL headers in our headers. The idea is that RetDec can internally use OpenSSL if it needs to, but it should not export it to its users. If we include OpenSSL in a public RetDec header, then we need to link OpenSSL publicly, which can cause a lot of trouble down the line. It would like to avoid it at all costs.

PeterMatula:
> e.g. not using these OpenSSL structures in headers at all

I did this in 741f4c39b58b063ba48ea73f061e8d26dfd69cef and 235228a06d933de592a5045b68ca9dd12bb23750. I'm closing this and hope that you won't have the compilation error anymore. If you do, please open an issue.

Issue #821, opened 2020-07-24
retdec-fileinfo crashes when detecting UPX


s3rvac:
`retdec-fileinfo` crashes when detecting UPX inside cpdetect.

**Steps to reproduce**

```
$ retdec-fileinfo FILE
```
where `FILE` is `0069BBD0F365ECF2020533B9DA5B702FABA808B22A05A3A42401A7065CBA263F`.

**Expected output**

Results from the analysis of the file.

**Actual output**

Segmentation fault.

**Output from valgrind**

```
Invalid read of size 8
   at 0x6FB52F: retdec::cpdetect::PeHeuristics::getUpxHeuristics() (pe_heuristics.cpp:940)
   by 0x70AB45: retdec::cpdetect::PeHeuristics::getFormatSpecificCompilerHeuristics() (pe_heuristics.cpp:2181)
   by 0x6ED119: retdec::cpdetect::Heuristics::getAllHeuristics() (heuristics.cpp:990)
   by 0x6D23B7: retdec::cpdetect::CompilerDetector::getAllHeuristics() (cpdetect.cpp:491)
   by 0x6D410D: retdec::cpdetect::CompilerDetector::getAllCompilers() (cpdetect.cpp:724)
   by 0x6D4718: retdec::cpdetect::CompilerDetector::getAllInformation() (cpdetect.cpp:793)
   by 0x44F37D: retdec::fileinfo::FileDetector::getCompilerInformation() (file_detector.cpp:77)
   by 0x450510: retdec::fileinfo::FileDetector::getAllInformation() (file_detector.cpp:337)
   by 0x4EDEB6: main (fileinfo.cpp:483)
 Address 0x0 is not stack'd, malloc'd or (recently) free'd
```

**Notes**

* The issue is in the following condition on line 940 in `src/cpdetect/heuristics/pe_heuristics.cpp`:
    ```cpp
    if (pos >= minPos && pos < 0x500 && pos < sections[0]->getOffset())
    ```
    For the above-mentioned file, `sections.size() == 0`, and the code crashes when accessing `sections[0]`. The `pos < sections[0]->getOffset())` condition was added in commit d8847281389d8e32b7f8cd49a94d7029d9c6a885.
* When fixing the issue, please add the file to our regression-tests suite.

**My configuration**

* Current `master` (235228a06d933de592a5045b68ca9dd12bb23750).
* 64b Debian Stretch, GCC 9, Debug build (however, I was also able to reproduce the issue with a release build)

PR #827, opened 2020-07-29
Fix crash on samples having corrupted PE header (#821)

tamaroth:
If a sample has corrupted PE header and as a result does not have any
sections, heuristic search for UPX packer will fail as it tries to
access the first section. To remedy this, we need to ensure the sections
exist before we access them.

Issue #821 

Test added [here](https://github.com/avast/retdec-regression-tests/pull/68)

### DIFF
--- a/src/cpdetect/heuristics/pe_heuristics.cpp
+++ b/src/cpdetect/heuristics/pe_heuristics.cpp
@@ -937,7 +937,7 @@ void PeHeuristics::getUpxHeuristics()
 	// format: x.xx'\0'UPX!
 	const std::size_t minPos = 5, verLen = 4;
 	pos = content.find("UPX!");
-	if (pos >= minPos && pos < 0x500 && pos < sections[0]->getOffset())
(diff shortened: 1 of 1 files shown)


src/cpdetect/heuristics/pe_heuristics.cpp
```diff
@@ -937,7 +937,7 @@ void PeHeuristics::getUpxHeuristics()
 	// format: x.xx'\0'UPX!
 	const std::size_t minPos = 5, verLen = 4;
 	pos = content.find("UPX!");
-	if (pos >= minPos && pos < 0x500 && pos < sections[0]->getOffset())
+	if (pos >= minPos && pos < 0x500 && sections.size() > 0 && pos < sections[0]->getOffset())
```
s3rvac, line 940: Consider replacing `sections.size() > 0` with `!sections.empty()` (we want to check if a vector is empty, which can be done directly using `std::vector::empty()`).

tamaroth: This is a good idea. Will fix it right away.

PR #834, opened 2020-08-05
Fixed build under RHEL based Linux where libraries are installed into lib64 instead of just lib

metthal:
When I tried to build RetDec on my Fedora 32, it didn't build bacause it was trying to look for yaramod in `lib` folder while RHEL based distros use `lib64` when 64-bit system is used. `CMAKE_INSTALL_LIBDIR` should take care of this.


PeterMatula:
Wow, I didn't know this. Few days ago I had a similar problem with Capstone being installed to to `lib` or `lib64` depending on the system. And I wrote some horrible hack to solve it :-D

PR #835, opened 2020-08-06
fix GCC 9.3.1 hang

hanickadot:
- under gcc 9.3.1 pe_dll_list's std::unordered_map causes infinite compilation time, I change it to make it compile in no time, and I'm using O(log_2 n) algorithm, which theoretically is slower than std::unordered_map but this will be quicker


PeterMatula:
run TC tests

PeterMatula:
TC test failed, but build is OK. This is most likely due to new tests in the `master`.

PR #848, opened 2020-09-08
deps/googletest: fix adding 'd' suffix in MSVC debug build

PeterMatula:


PeterMatula:
lets run teamcity tests

PR #853, opened 2020-09-17
Added YARA rules for FlyStudio installer

ladislav-zezula:


s3rvac (approved):
Thanks! :+1: Fur future reference, tests are in https://github.com/avast/retdec-regression-tests/pull/71.

PR #857, opened 2020-09-22
GhostInstaller and InnoSetup revisited

ladislav-zezula:
Regression tests: https://github.com/avast/retdec-regression-tests/pull/73


s3rvac (approved):
Thanks! :+1:

PR #858, opened 2020-09-23
Lz installers install creator

ladislav-zezula:
Tests: https://github.com/avast/retdec-regression-tests/pull/74


s3rvac (approved):
Thanks! :+1:

PR #865, opened 2020-10-06
improved YARA rules for Wise installer

ladislav-zezula:
Tests: https://github.com/avast/retdec-regression-tests/pull/76


s3rvac (approved):
Thanks! :+1:

PR #894, opened 2020-11-14
Fix missing comma in usualPackerSections initializer

HoundThe:
While I was working on Signatures, I've noticed that usualPackerSections set is missing a comma in it's initializer, I think this is not intended?


s3rvac (approved):
Yes, definitely a missing comma. Thank you!

PR #897, opened 2020-11-23
Lz import thunk check

ladislav-zezula:


s3rvac (approved):
Tests are passing (apart from the [macOS build](https://retdec-tc.avast.com/viewLog.html?buildId=5694&buildTypeId=Retdec_MacBuild&tab=buildLog), which has timeouted twice during installation), and I see no issues with the code, so I am going to merge the PR.

Thanks for the fix.

PR #913, opened 2021-01-07
unpacker/mpress: Properly copy non-packer related sections to the unpacked file

metthal:
After the commit 60ea783b558cce9b1f1a419df9c75be15588f3d2, we've had MPRESS unpacker bugged. Specifically the part which copied sections from packed PE into unpacked PE, which were not modified by packing process at all. This resulted in files unpacked by our `retdec-unpacker` to miss things like exports, resources and other things which are usually placed in their own sections (apart from imports which were properly reconstructed and unpacked code).


metthal:
Lets run TC tests.

PR #936, opened 2021-03-20
Implement telfhash for ELF import table

HoundThe:
I've added tlsh https://github.com/trendmicro/tlsh code into the project and used it to calculate telfhash for ELF import tables.


include/retdec/fileformat/utils/tlsh/tlsh_impl.h
```diff
@@ -0,0 +1,164 @@
+/*
```
s3rvac, line 1: For @PeterMatula to consider: Do we really want a verbatim copy of [tlsh](https://github.com/trendmicro/tlsh) in our repo? Other dependencies are pulled through external projects. Now, it looks as if tlsh was part of RetDec (AFAIK, we use `src` and `include` for our own code and `deps` for dependencies). If you have agreed with @HoundThe on that, then OK, but I just wanted to be sure.

PeterMatula: Since adding more external deps is PITA, and tlsh is only a few C++ sources, we agreed on inlining them to our repo. But definitely not at their current location. They should go into `/deps/` as other external projects. I will for sure mention this in my review.

PeterMatula:
Lets run TC tests.

include/retdec/fileformat/types/import_table/elf_import_table.h
```diff
@@ -0,0 +1,18 @@
...
+namespace retdec {
+namespace fileformat {
+
+class ElfImportTable : public ImportTable
+{
+public:
+	void computeHashes() override;
+};
+} // namespace fileformat
+} // namespace retdec
```
PeterMatula, line 18: No new line at the end of file.

include/retdec/fileformat/types/import_table/import_table.h
```diff
@@ -11,7 +11,6 @@
 #include <vector>
 
 #include "retdec/fileformat/types/import_table/import.h"
-
```
PeterMatula, line 14: why remove the line?

src/fileformat/types/import_table/elf_import_table.cpp
```diff
@@ -0,0 +1,96 @@
...
+#include "retdec/fileformat/types/import_table/elf_import_table.h"
+#include <algorithm>
+#include <regex>
+
+using namespace retdec::utils;
+
+namespace retdec {
+namespace fileformat {
+
+static const std::set<std::string> exclusion_set = {
```
PeterMatula, line 19: Where did the list come from? Is this the same and complete list as in the original [telfhash](https://github.com/trendmicro/telfhash) implementation?  Some reference to the original or explanation why and what is on the list would be useful if somebody wanted to update it in the future.

PeterMatula: Also, `std::unordered_set`?

src/fileformat/types/import_table/elf_import_table.cpp
```diff
@@ -0,0 +1,96 @@
...
+	"main", // main function
+	"abort", // ARM default
+	"cachectl", // MIPS default
+	"cacheflush", // MIPS default
+	"puts", // Compiler optimization (function replacement)
+	"atol", // Compiler optimization (function replacement)
+	"malloc_trim" // GNU extensions
+};
+
+bool isSymbolExcluded(std::string symbol)
```
PeterMatula, line 30: Why pass string as value? Are you modifying it in the function? If not then `const std::string&` or `std::string_view`.

src/fileformat/types/import_table/elf_import_table.cpp
```diff
@@ -0,0 +1,96 @@
...
+	if (symbol.empty()) {
+		return true;
+	}
+	/* ignore:
+		symbols starting with . or _
+		x86-64 specific functions
+		string functions (str.* and mem.*), gcc changes them depending on architecture
+		symbols starting with . or _
+	*/
+	std::regex exclusion_regex("(^[_\.].*$)|(^.*64$)|(^str.*$)|(^mem.*$)");
```
PeterMatula, line 41: Again, how/why/where was this get? What if somebody wanted to modify it?

src/fileformat/types/import_table/elf_import_table.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+	if (std::regex_match(symbol, exclusion_regex)) {
+		return true;
+	}
+
+	if (exclusion_set.count(symbol)) {
+		return true;
+	}
+
+	return false;
```
PeterMatula, line 51: Looks like the whole function could be simplified to something like:
```
return symbol.empty()
    || std::regex_match(symbol, exclusion_regex)
    || exclusion_set.count(symbol);
```

src/fileformat/types/import_table/elf_import_table.cpp
PeterMatula: Also, with regexes one always have to think when and how many times they get compiled. I always have to google it myself, so I will leave this one for you - read on it and think about it. Maybe I'm wrong, but I think this regexe could be made `static`.

src/fileformat/types/import_table/elf_import_table.cpp
```diff
@@ -0,0 +1,96 @@
...
+}
+
+/**
+ * Compute import telfhash
+ */
+void ElfImportTable::computeHashes()
+{
+	std::vector<std::string> imported_symbols;
```
PeterMatula, line 59: You could preallocate vector using `vector::reserve(imports.size())`. Since you are filtering some out, but not adding, this should safely over-approximate the space needed. It is a good habit with vectors if you know the size beforhand.

src/fileformat/types/import_table/elf_import_table.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+		// filter the symbols just as telfhash does for the same result
+		if (isSymbolExcluded(funcName)) {
+			continue;
+		}
+
+		imported_symbols.push_back(funcName);
+	}
+
+	std::sort(imported_symbols.begin(), imported_symbols.end());
```
PeterMatula, line 72: If you are sorting, why not use `std::set`? Are duplicates possible/desirable? If yes, and we want to keep them, `std::vector` is ok. If we don't want them, `std::set` would solve it for us.
Does telfhash  allow duplicates in this container?

PeterMatula: Also a thing, is the sorting algorithm behaving the same as in the telfhash implementation? I think it is possible to somehow control the string sorting in C++? There might be differences. And even it there are not, a comment on this would be helpful so that no one screws it up later.

src/fileformat/types/import_table/elf_import_table.cpp
```diff
@@ -0,0 +1,96 @@
...
+	for (const auto& symbol : imported_symbols) {
+		if (!impHashString.empty())
+			impHashString.append(1, ',');
+
+		impHashString.append(symbol);
+	}
+
+	if (impHashString.size()) {
+		Tlsh tlsh;
+		tlsh.update(reinterpret_cast<const uint8_t*>(impHashString.data()), impHashString.size());
```
PeterMatula, line 84: Please no lines over 80 chars long in the newly added code :-)
I know, there is a shitload of them in the old code... and I hate them all.
Set some red border in the editor.

src/fileformat/types/import_table/elf_import_table.cpp
```diff
@@ -0,0 +1,96 @@
...
+		impHashTlsh = toLower(tlsh.getHash(show_version));
+
+		impHashCrc32 = getCrc32(reinterpret_cast<const uint8_t*>(impHashString.data()), impHashString.size());
+		impHashMd5 = getMd5(reinterpret_cast<const uint8_t*>(impHashString.data()), impHashString.size());
+		impHashSha256 = getSha256(reinterpret_cast<const uint8_t*>(impHashString.data()), impHashString.size());
+	}
+}
+
+} // namespace fileformat
+} // namespace retdec
```
PeterMatula, line 96: no new line

src/fileformat/types/import_table/import_table.cpp
```diff
@@ -675,6 +675,14 @@ const std::string& ImportTable::getImphashSha256() const
 	return impHashSha256;
 }
 
+/**
+ * Get imphash as tlsh
+ * @return Imphash as tlsh
+ */
```
PeterMatula, line 681: You don't have to write comments like these. Ideally functions should be clear from their signature - name, return type, argument types and names. This one is. The comment is not adding any useful new information.

PeterMatula (changes requested):
* `linux-doxygen-build` failed.
* `win-build` failed - looks like it has trouble compiling TLSH.
* TLSH sources must not be between other RetDec sources. Move it to `/deps/tlsh`. Create CMake library target for it - look for example at `tinyxml2` which should be similar. I can help with CMake if needed. Also, add at least license file to sources, and some version identification, e.g. version file if it exists - preferably use some tagged TLSH version, not random commit at `master`. Also, add TLSH entry to our `/LICENSE-THIRD-PARTY` file.
* I also expect PR in [retdec-regression-tests](https://github.com/avast/retdec-regression-tests) that on some real-world input sample tests both plain and json values. Place it in `/tools/fileinfo/features`.
* Can you send me a file, where it produces some hash? I tried it on `retdec-regression-tests/integration/ack/2018-09-13/ack.x64.gcc.O0.g.elf` and I got newly added entries for `SHA256`, `MD5`, `CRC32`, but no TLSH. Is it working for you? M'I doing something wrong? This is exactly the thing that regression tests would solve automatically.
* After you fix these issues, and test the implementation on those 450k samples I sent, change the PR from draft to a full PR, and let me know. I will go over it again and merge it.

src/fileformat/types/import_table/elf_import_table.cpp
HoundThe: Original implementation seems to use a python list and not check for duplicates, so I think it allows duplicates. Python sort and C++ one on strings should sort lexicographically. But I'll be more sure after further testing.

HoundThe:
> * Can you send me a file, where it produces some hash? I tried it on `retdec-regression-tests/integration/ack/2018-09-13/ack.x64.gcc.O0.g.elf` and I got newly added entries for `SHA256`, `MD5`, `CRC32`, but no TLSH. Is it working for you? M'I doing something wrong? This is exactly the thing that regression tests would solve automatically.

This one for example https://www.virustotal.com/gui/file/fb5b5659801beccc790444f5fecb6acc74c10469acf7554f8f1eeea033bbbfe8
I think the file you tested it on has too few symbols to generate the hash from. I am not sure what is the exact minimum input, from the tlsh source it seems to be this `#define MIN_DATA_LENGTH              50`, the binary you've tried has data length of only 6, as the only non-filtred import symbol is printf. Can be tested by the original telfhash implementation, which outputs tnull:

```
$ telfhash retdec-regression-tests/integration/ack/2018-09-13/ack.x64.gcc.O0.g.elf -d
ELFCLASS64
8 symbols found
1 symbols considered:
printf
retdec-regression-tests/integration/ack/2018-09-13/ack.x64.gcc.O0.g.elf  tnull
```

PeterMatula:
Lets run TC tests.

HoundThe:
I've moved TLSH to deps and wrote a CMakeList for it inspired by existing ones, but it's the first time I used CMake so I am not sure if I did it correctly, especially with the cond_add  which I've set if RETDEC_ENABLE_FILEFORMAT is set. 

I've added their license file content into the LICENSE-THIRD-PARTY file. But I am not sure what should I put into the license file and version file that I should put among the source files. I've used the latest tagged release https://github.com/trendmicro/tlsh/releases/tag/4.2.1 . The version values are in the CMakeList

HoundThe:
After testing out the implementation on 464527 ELF binaries I've found few things about telfhash that I didn't really account for:

- telfhash hashes also exported symbols, as in symbols that have value.
- telfhash cares only about sections, if the binary has symbols in dynamic segments, telfhash doesn't use them even tho RetDec could because it's capable to extract them.

The first one results in a situation when we have a telfhash, but because there are no imported symbols, there is no import table in the output. Maybe use something else strictly for import table and use telfhash separately? That could also give ability to use symbols from the dynamic segments.

HoundThe:
Lets run TC tests.

HoundThe:
I moved the telfhash to the `ElfFormat` and kept the existing implementation in import table as a general import hash. I've also added TLSH hash to every import table, because I thought why not, but I can remove it. 

I've tested it and seems to work well compared to original implementation, I think it's ready for re-review.

include/retdec/fileformat/file_format/elf/elf_format.h
```diff
@@ -13,6 +13,7 @@
 
 #include "retdec/fileformat/file_format/file_format.h"
 #include "retdec/fileformat/types/note_section/elf_notes.h"
+#include "retdec/fileformat/types/import_table/elf_import_table.h"
```
PeterMatula, line 16: Does this have to be included here? If not, it is best to hide includes in cpp files.

include/retdec/fileformat/file_format/elf/elf_format.h
```diff
@@ -23,6 +24,11 @@ namespace fileformat {
 class ElfFormat : public FileFormat
 {
 	private:
+		std::vector<std::string> telfhashSymbols;
```
PeterMatula, line 27: Does this have to be in the header? It would be best to have it only locally in source and pass it to `loadTelfhash()` fuction.

include/retdec/fileformat/file_format/elf/elf_format.h
```diff
@@ -23,6 +24,11 @@ namespace fileformat {
...
+		/// flag if we already loaded symbols from SHT_DYNSYM
+		bool telfhashDynsym = false;
```
PeterMatula, line 29: Again, could we maybe get rid of this somehow? If not then ok, but these flags are often hard to maintain, it would be better to determine this somehow from the result data themselves, or something else.

src/fileformat/types/import_table/import_table.cpp
```diff
@@ -820,11 +825,19 @@ void ImportTable::computeHashes()
...
 
-	if (impHashBytes.size())
-	{
-		impHashCrc32 = getCrc32((const uint8_t *)impHashBytes.data(), impHashBytes.size());
-		impHashMd5 = getMd5((const uint8_t *)impHashBytes.data(), impHashBytes.size());
-		impHashSha256 = getSha256((const uint8_t *)impHashBytes.data(), impHashBytes.size());
+	if (impHashBytes.size()) {
+		auto data = reinterpret_cast<const uint8_t*>(impHashBytes.data());
+
+		Tlsh tlsh;
```
PeterMatula, line 831: As I looked into these hashes for this review, I realized this entire `import_table.cpp` implementation is just for PE. It does not work for ELF at all. Once we merge this PR, I will refactor it to some PE-specific module. As such I don't think telfhash is needed here, it just confuses the thing even more -> remove its computation here.

HoundThe: This import table is used also by Mach-O.

include/retdec/fileformat/file_format/elf/elf_format.h
HoundThe: I didn't find a nicer way to keep the state, because of the way the symbol loading works, I must hold it through multiple functions calls of `loadSymbols()`. I guess static variable in the function works too, but I am not sure if that is a better solution tho.

Issue #621, opened 2019-08-07
Generate hash out of rich header in fileformat


metthal:
Rich headers are long in length and we use them frequently. Generating hash out of the unmasked rich header data could be useful to reduce this length. The hash would be generated in the same way as we do it with import table, export table or any other.

Presentation in fileinfo should be also done in this ticket.

PeterMatula:
Implemented in #945.

PR #945, opened 2021-04-14
Calculate Rich header hash

HoundThe:
I've added a calculation of decrypted (XORed) rich header hashes (crc32, md5, sha256) as mentioned in issue #621 . Not sure if all of them are wanted, probably MD5 only would be ok as used by VirusTotal: https://www.virustotal.com/gui/file/0ee196bb23f0eafe4f61d30bf6c676fd7365cb12ae66a6bde278851e91901ac1/details


PeterMatula (approved):
Meergin this, but @HoundThe please also add [regression tests](https://github.com/avast/retdec-regression-tests) checking all three values - on some sample where the results is the same as in Virus Total.

PeterMatula:
@HoundThe I will add the regression test myself, it will take just a minute and we can consider this done.

PR #970, opened 2021-06-15
Fixed Lz issue 967

ladislav-zezula:
.NET Edge cases are now handled properly


PeterMatula (approved):
As far as I can say, it looks ok :+1: (but to be fair, this is too low-level for me to know what exactly you are doing :-D)

ladislav-zezula:
> but to be fair, this is too low-level for me to know what exactly you are doing :-D

I'll gladly explain it to you on a call :-)

PR #973, opened 2021-06-21
Fix the Rich header analysis algorithm

### LINKED ISSUE #960: Fileinfo: junk in PE rich header signature
We encountered junk data provided by `retdec-fileinfo` in PE rich header signature. E.g. for `2acd2ff9c70ba9398221cf2265b2fddaceae3e31a29883594bcce545f02be6a3`:
```
Rich header signature    : 89de35fc89de359489de35b789de355989de355489de352689de34dd89de34a889de34b289de346589de341e
                           89de342289de343b89de336789de330389de332989de32c189de328289de32a389de324989de325189de3267
...
```
Samples:
* 2acd2ff9c70ba9398221cf2265b2fddaceae3e31a29883594bcce545f02be6a3
* 7f29a26f830eee42a80a1a35169d9f616ca9823e386316f5eccfe36f90a8fe4b

Investigate the reasons and try to prevent providing such junk data. The solution would be to either fix a bug causing this (if there is a bug) or reliably detect such cases and prevent them. It is however quite possible that file offsets are in fact pointing at existing "junk" data, try to come up with some solution anyway - e.g. heuristics, sanity checks - analyze and discuss with @PeterMatula.

### LINKED ISSUE #965: Detection of rich header offset should also check for valid Rich magic
Right now, valid rich magic `Rich` is not checked for when detecting rich header offset. In samples like `646679c8decab9e85874c306560e75df410013bf9de2aed17a3e92aedc7c6dad` it causes issues because it detects rich header on lower offset than it really is. The real offset of rich header should be 0x90 but we detect it as 0x88.

HoundThe:
This PR aims to fix #960 and to fix #965
The Rich header end is marked by the "Rich" marker somewhere between DOS header/stub and PE header. After the "Rich" follows the XOR key, when XORing the Rich header data, the start can be identified by "DanS" in ASCII and 3 null bytes for padding into a 16-byte paragraph. 

Current analysis decrypts the whole thing as it assumes there is no other data than Rich header there and then checks if there is "DanS" at the start. This doesn't behave correctly if there is any other "junk" data before the Rich header as it will decrypt it as well as is the case in #960 and just reports warning about wrong key because it didn't find the "DanS" at the start. I've changed the algorithm so it starts decrypting from the end ("Rich") to the top until it finds the "DanS" that marks the start of the Rich header, this seems to work well for both samples 

 - 2acd2ff9c70ba9398221cf2265b2fddaceae3e31a29883594bcce545f02be6a3
 - 7f29a26f830eee42a80a1a35169d9f616ca9823e386316f5eccfe36f90a8fe4b

But this means that the Rich header offset can vary based on the junk data, but this isn't really supported by the current RetDec code structure, so I'll need to edit that so we return the correct offset.

Checklist for PR:
- [x] Fix the analysis algorithm
- [x] Propagate correct offset from the Rich header analysis
- [x] Add test case into the regression test suite

### DIFF
--- a/include/retdec/pelib/RichHeader.h
+++ b/include/retdec/pelib/RichHeader.h
@@ -24,7 +24,7 @@ namespace PeLib
 			bool headerIsValid;
 			bool validStructure;
 			std::uint32_t key;
-			std::uint64_t offset;
+			std::uint64_t offset = 0;
 			std::size_t noOfIters;
 			std::vector<std::uint32_t> decryptedHeader;
 			std::vector<PELIB_IMAGE_RICH_HEADER_RECORD> records;
--- a/include/retdec/pelib/RichHeader.h
+++ b/include/retdec/pelib/RichHeader.h
@@ -24,6 +24,7 @@ namespace PeLib
 			bool headerIsValid;
 			bool validStructure;
 			std::uint32_t key;
+			std::uint64_t offset;
 			std::size_t noOfIters;
 			std::vector<std::uint32_t> decryptedHeader;
 			std::vector<PELIB_IMAGE_RICH_HEADER_RECORD> records;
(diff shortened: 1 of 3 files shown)


HoundThe:
This PR is also solution to #965

PeterMatula (approved):

PR #978, opened 2021-06-25
Increased PE symbols MAX_LENGTH limits

### LINKED ISSUE #957: Bump MAX_LENGTH constants in pelib to reflect what we see in the wild
We've ran into some samples which contain very long symbols which can be considered valid. There are however several `MAX_LENGTH` constants in PeLib which limit them to 96. It'd be the best to push these constants further so that we can realistically parse out what is possible but still with reasonable constraints (let's say 256, 384 or 512 in the worst case). They are all located in `include/retdec/pelib/PeLibAux.h` but try to look through the whole `pelib` folder to find other limit constants.

Sample for test: `1c9620c688d00703a5c535ffbd59f27f5f7da07b18ccb12aef480ac2c0222324`

HoundThe:
Fixes #957 

It seems there are 3 constants in the header:
- IMPORT_LIBRARY_MAX_LENGTH
- IMPORT_SYMBOL_MAX_LENGTH
- COFF_SYMBOL_NAME_MAX_LENGTH

I was wondering why isn't there a constant for exported symbols. It seems that for exported symbols there is no "sanity" length cap, it's capped by the default max value of the `read()` function (which is used to read the string from the image) which is 65535. I haven't seen any other MAX length constants in the pelib headers.

Based on the example binary I've raised those 3 constants to 256.

### DIFF
--- a/include/retdec/pelib/PeLibAux.h
+++ b/include/retdec/pelib/PeLibAux.h
@@ -1127,8 +1127,8 @@ namespace PeLib
 		}
 	};
 
-	const std::size_t IMPORT_LIBRARY_MAX_LENGTH = 96;
-	const std::size_t IMPORT_SYMBOL_MAX_LENGTH = 96;
+	const std::size_t IMPORT_LIBRARY_MAX_LENGTH = 256;
+	const std::size_t IMPORT_SYMBOL_MAX_LENGTH = 256;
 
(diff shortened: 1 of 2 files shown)


PeterMatula:
Merging this, but I think exports could/should be capped as well.  What does @metthal think? We could create another issue for this...

PeterMatula (approved):

PR #979, opened 2021-06-25
Change the section name parsing to only remove trailing zeroes

### LINKED ISSUE #958: Improve PE section name handling of escaped sequences
Files like `0043812838495a45449a0ac61a81b9c16eddca1ad249fb4f7fdb1c4505e9bb34` contain sections which do not abide specs of PE format. It contains section names with null terminators in them. Fileinfo is somehow able to handle these but they are not correctly printed out because instead of `.CRT\x00*ls` we only get `.CRT*ls`.

Please try to resolve it so that section name is correctly printed with escaped non-printable characters and rightmost zero padding is removed (basically `rstrip` of `\x00`).

HoundThe:
Fixes #958 

The removal of all NULL chars in the pelib was intentional as in the comment: 
```
	// Historically, PELIB copies the name of the section WITHOUT zero chars,
	// even if the zero chars are in the middle. Aka ".text\0\0X" results in ".textX"
```
Not sure what was the reason, but based on the issue, I've changed the parsing to strip only the trailing zeroes.

### DIFF
--- a/src/pelib/ImageLoader.cpp
+++ b/src/pelib/ImageLoader.cpp
@@ -4,6 +4,7 @@
  * @copyright (c) 2020 Avast Software, licensed under the MIT license
  */
 
+#include <algorithm>
 #include <iostream>
 #include <fstream>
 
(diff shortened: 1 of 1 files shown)


PeterMatula (commented):
These tricky iterations again :-D
Corner cases:
1. I think it can handle even the case when an entire name is `\0`. But do we have this in regression tests? I think not. Can we add it? The easiest would probably be to hack some sample, possible the one already used if there are some more sections to play with, with some hex editor.
2. Can we handle if the name is not terminated at all - all the chars are printable? I think we decrease `end` once anyway, and then assign `end + 1` as end. So it is probably ok. But again, can we have a test for this - if there already isn't one?

HoundThe:
I think I tried all of these corner cases and I still have the samples I've created for this. I'll have a look into the regression suite if there is something like that, if not then I'll create a single sample that has these cases and update the tests.

HoundThe:
I've added the test cases

PR #981, opened 2021-06-28
Lz retdec 54

ladislav-zezula:


PeterMatula:
Let's run TC tests.

PeterMatula (approved):

PeterMatula:
@ladislav-zezula Code and tests look ok, but `features.windrivers.Test` regression tests is failing across all the architectures. I hoped it is some small detail, but it looks like after this change, there are major differences in the underlying data. I'm adding outputs before and after the change. There is much less code in the output C after the change. In `.dsm` files, we can see that before the change, there were much more data that got decoded into instructions. After the change, these bytes are zero and we don't get instruction, functions, etc.
I'm not sure if this is ok - i.e. it was bad before, or if your changes have some bad side effects. Can you please have a look - you know more about it than I do. If you won't know right away, we may investigate it together.
[out.zip](https://github.com/avast/retdec/files/6847223/out.zip)

ladislav-zezula:
Let's run TC tests.

Issue #963, opened 2021-06-06
Investigate discrepancies in PE resource detection compared to YARA


metthal:
Sometimes we provide different PE resources than YARA. I haven't dig into this much deeper and don't know the cause but it should be investigated and fixed afterwards.

Files:

```
88ff9ef1f056be47f0da0d226e1295ba7a9668595dd5af5da3fcce89eafc204d
8a8154c36977c3f47a65c329a6dd93c58b8aff6050a4a748f4c610310daf0d8b
```

HoundThe:
Hey, I wanna ask about the second file `8a8154c36977c3f47a65c329a6dd93c58b8aff6050a4a748f4c610310daf0d8b` Is there some specific thing that differs? I've tried to dump module data from YARA and output from FileInfo, both found the same amount of resources and on first sight, few first and last entries seem to be about the same.

The first file `88ff9ef1f056be47f0da0d226e1295ba7a9668595dd5af5da3fcce89eafc204d` has 224 resources in FileInfo and 0 in YARA, so I'll start with that one.

HoundThe:
The first file seems to be ignored by YARA because the resource entries contain non-null characteristics. (thanks to @ ladislav-zezula)
After a deeper look, for the second file, YARA returns YR_UNDEFINED as offset value at certain entries, probably a sanity check for the offset validity.

HoundThe:
I've created PR that addresses the second sample with invalid offsets. Based on discussion with @ ladislav-zezula the first sample being different in YARA is wrong on the YARA side and not FileInfo.

PeterMatula:
Second case fixed in https://github.com/avast/retdec/pull/982.
As the first case doesn't seem to be a RetDec problem I'm closing this for now.

@metthal please have a look and do one of these:
1. If you don't agree, and think this should be addressed in RetDec please reopen this.
2. If you think this should be addressed in YARA, either create an issue there or let me know, I will do it.
3. Ignore.

PR #982, opened 2021-07-05
Add check that the resource file offset is valid

HoundThe:
Addresses partially #963 

Add check that the file offset of a Resource is actually somewhere inside section physical data

Sample that has such invalid offsets: 
`8a8154c36977c3f47a65c329a6dd93c58b8aff6050a4a748f4c610310daf0d8b`

If this fix is wanted, there are few things to consider:
 - This means that offset 0 is considered bad. If 0 offsets could be valid (But I think they are not?) then the feedback about invalid offset should be changed
 - For the offset 0 entries, the hash is calculated, which is unnecessary and could be misleading. This needs to be addressed too before possible merging.

### DIFF
--- a/src/fileformat/types/resource_table/resource.cpp
+++ b/src/fileformat/types/resource_table/resource.cpp
@@ -344,7 +344,7 @@ void Resource::setSublanguageId(std::size_t rId)
  */
 bool Resource::isValidOffset() const
 {
-	return this->offset != UINT32_MAX ? true : false;
+	return offset != UINT32_MAX;
 }
 
 /**
--- a/src/fileformat/types/resource_table/resource.cpp
+++ b/src/fileformat/types/resource_table/resource.cpp
@@ -337,6 +337,17 @@ void Resource::setSublanguageId(std::size_t rId)
 }
 
 /**
+ * @brief Checks if the offset is valid, with UINT32_MAX reserved as invalid
+ * 
+ * @return true 
+ * @return false 
+ */
+bool Resource::isValidOffset() const
+{
+	return this->offset != UINT32_MAX ? true : false;
+}
+
+/**
  * A method which indicates whether resource is loaded.
  * @return `true` if it is, otherwise `false`.
  */
--- a/src/pelib/ImageLoader.cpp
+++ b/src/pelib/ImageLoader.cpp
@@ -483,6 +483,50 @@ uint32_t PeLib::ImageLoader::getFileOffsetFromRva(uint32_t rva) const
 	return rva;
 }
 
+// similar to getFileOffsetFromRva, but the offset is within the real file and not memory image
+uint32_t PeLib::ImageLoader::getValidOffsetFromRva(uint32_t rva) const
+{
+	// If we have sections loaded, then we calculate the file offset from section headers
+	if (sections.size())
+	{
+		// Check whether the rva goes into any section
+		for (auto& sectHdr : sections)
+		{
+			// Only if the pointer to raw data is not zero
+			if (sectHdr.PointerToRawData != 0 && sectHdr.SizeOfRawData != 0)
+			{
+				uint32_t realPointerToRawData = sectHdr.PointerToRawData;
+				uint32_t sectionRvaStart = sectHdr.VirtualAddress;
+				uint32_t virtSize = sectHdr.VirtualSize;
+				uint32_t rawSize = sectHdr.SizeOfRawData;
+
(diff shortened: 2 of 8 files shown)


metthal:
> For the offset 0 entries, the hash is calculated, which is unnecessary and could be misleading. This needs to be addressed too before possible merging.

Good point, could this be please incorporated into these changes?

HoundThe:
I've added a flag that represents the offset validity, so invalid offsets are not exported to the output at all.

src/pelib/ImageLoader.cpp
```diff
@@ -486,6 +486,50 @@ uint32_t PeLib::ImageLoader::getFileOffsetFromRva(uint32_t rva) const
 	return rva;
 }
 
+// similar to getFileOffsetFromRva, but the offset is within the real file and not memory image
+uint32_t PeLib::ImageLoader::getValidOffsetFromRva(uint32_t rva) const
```
metthal, line 490: Please use `std::uint32_t` instead of just `uint32_t` to follow used standards in retdec.

src/pelib/ImageLoader.cpp
```diff
@@ -486,6 +486,50 @@ uint32_t PeLib::ImageLoader::getFileOffsetFromRva(uint32_t rva) const
...
+				{
+					// Make sure we round the pointer to raw data down to PELIB_SECTOR_SIZE.
+					// In case when PointerToRawData is less than 0x200, it maps to the header!
+					return realPointerToRawData + offset;
+				}
+			}
+		}
+
+		// Check if the rva goes into the header
+		return (rva < optionalHeader.SizeOfHeaders) ? rva : UINT32_MAX;
```
metthal, line 527: Instead of `UINT32_MAX` being reserved as invalid value, what about going more C++ style and use `std::optional<std::uint32_t>`?

src/fileformat/types/resource_table/resource.cpp
```diff
@@ -336,6 +336,17 @@ void Resource::setSublanguageId(std::size_t rId)
...
 
+/**
+ * @brief Checks if the offset is valid, with UINT32_MAX reserved as invalid
+ * 
+ * @return true 
+ * @return false 
+ */
+bool Resource::isValidOffset() const
+{
+	return this->offset != UINT32_MAX ? true : false;
```
metthal, line 347: Is this `this` really necessary here? Also condition `<cond> ? true : false` can be simplified just as `<cond>`.

metthal (changes requested):

src/pelib/ImageLoader.cpp
metthal: I haven't noticed this is pelib code which might use different code style, in that case disregard.

HoundThe: Yeah, ImageLoader has them typedef'd at the top of source file with comment `// Anti-headache` so I just followed the used style within the unit

src/pelib/ImageLoader.cpp
HoundThe: I guess that's possible, I was following the style of a similar function like `getFileOffsetFromRva(uint32_t rva)` Which is close to identical and uses this representation of invalid value. I can change it, but it might look inconsistent with the existing code and lead to confusion?

src/fileformat/types/resource_table/resource.cpp
HoundThe: Thanks, fixed!

src/pelib/ImageLoader.cpp
HoundThe: Also regarding coding styles, just a thought - wouldn't it be better to unify the coding styles within the RetDec with some `.clang-format` rules?

PeterMatula: Please remove those `Anti-headache` defs, lets use `std::whatever` everywhere.

PeterMatula:
lets run TC tests

src/pelib/ImageLoader.cpp
HoundThe: Done

src/pelib/ImageLoader.cpp
metthal: Yeah, I haven't noticed that it's in many other functions and would generally require many changes that are out of scope of this PR. Let's just do it this way for now.

PeterMatula (approved):

PR #987, opened 2021-07-15
Fix UB due to iterator reaching before begin() in rich header parsing 

HoundThe:
There was an undefined behavior because iterator pointing to begin() got decremented when the rich header parsing loop ends.

### DIFF
--- a/src/pelib/RichHeader.cpp
+++ b/src/pelib/RichHeader.cpp
@@ -882,15 +882,15 @@ namespace
 
 			// Start analyzing from the end - "Rich" marker
 			// and move upwards to decrypted "DanS" marker
-			for (auto i = richSignature - 1; i >= rich.begin(); --i)
+			for (auto it = std::make_reverse_iterator(richSignature); it < rich.rend(); it++)
 			{
-				std::uint32_t decrypted_dword = *i ^ key;
+				std::uint32_t decrypted_dword = *it ^ key;
 				decryptedHeader.push_back(decrypted_dword);
 				// "DanS" - 0x536e6144 signals the start (end) of the rich header
 				if (decrypted_dword == 0x536e6144)
 				{
 					// Set the offset to "DanS"
-					this->offset = (i - rich.begin()) * 4;
+					this->offset = std::distance(it + 1, rich.rend()) * 4;
 					// Because we are analysing bottom up, reverse the vector
 					std::reverse(decryptedHeader.begin(), decryptedHeader.end());
 					break;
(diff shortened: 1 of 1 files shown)


src/pelib/RichHeader.cpp
```diff
@@ -882,15 +882,15 @@ namespace
...
 			{
-				std::uint32_t decrypted_dword = *i ^ key;
+				std::uint32_t decrypted_dword = *it ^ key;
 				decryptedHeader.push_back(decrypted_dword);
 				// "DanS" - 0x536e6144 signals the start (end) of the rich header
 				if (decrypted_dword == 0x536e6144)
 				{
 					// Set the offset to "DanS"
-					this->offset = (i - rich.begin()) * 4;
+					this->offset = std::distance(it + 1, rich.rend()) * 4;
```
PeterMatula, line 893: This iterator arithmetic is always a bit tricky:
1) Is that `it + 1` always valid in this case? Cannot be again get to some trouble if be are at corner cases?
2) Couldn't we get the same result, but move the addition away from the iterator?
```
std::distance(it, rich.rend()) + 1
```
or would at actually be this, because they are reverse iterators?
```
std::distance(it, rich.rend()) - 1
```
and if so, couldn't we get to `0 - 1` which would be a problem? Keep in mind that `rich.rend()` is one before the first.

PeterMatula (commented):
I think it is ok, but (reverse) iterators are tricky, so please give it a second look based on my comments.

src/pelib/RichHeader.cpp
HoundThe: I gave it some extra thought before I've pushed this, my thoughts:
The loop doesn't get executed when `it == rich.rend()`, then `it` is always pointing to one element greater. Thus corner case is if we are pointing to the first element (index 0), then `it + 1 == rich.rend()`, which results at 0 distance, and 0 offset, as we're pointing to the first element and so there is no offset.

Your example `std::distance(it, rich.rend()) - 1` is also an option, I was thinking the distance is more readable so I went with that, but I can change it if it's more clear that way.
> "and if so, couldn't we get to 0 - 1 which would be a problem?"

We couldn't because in the loop, distance between `it` and `rich.rend()` will never be 0 (which is guaranteed by the for loop condition), which is the reason for the + 1 I have in there.

PeterMatula: Great, I came to the same conclusion.

src/pelib/RichHeader.cpp
```diff
@@ -882,15 +882,15 @@ namespace
 
 			// Start analyzing from the end - "Rich" marker
 			// and move upwards to decrypted "DanS" marker
-			for (auto i = richSignature - 1; i >= rich.begin(); --i)
+			for (auto it = std::make_reverse_iterator(richSignature); it < rich.rend(); it++)
```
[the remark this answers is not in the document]
PeterMatula: i fixed it

PR #994, opened 2021-07-29
Add signatureVerified flag for each signature

HoundThe:
Add a flag for each signature that represents if the signature and its signer were successfully verified (digest matches etc.) in the same manner as the previous implementation.

### DIFF
--- a/src/fileformat/file_format/pe/authenticode/pkcs7_signature.cpp
+++ b/src/fileformat/file_format/pe/authenticode/pkcs7_signature.cpp
@@ -560,11 +560,7 @@ std::vector<DigitalSignature> Pkcs7Signature::getSignatures(const retdec::filefo
 	// If there are warnings, then the signature is invalid
 	// We don't use the countersignature warnings here as
 	// previous implementation did when verifying
-	if (signature.warnings.empty()) {
-		signature.isValid = true;
-	} else {
-		signature.isValid = false;
-	}
+	signature.isValid = signature.warnings.empty();
 
 	for (auto&& counterSig : signInfo.counterSignatures) {
 		CertificateProcessor processor;
--- a/src/fileformat/file_format/pe/authenticode/pkcs7_signature.cpp
+++ b/src/fileformat/file_format/pe/authenticode/pkcs7_signature.cpp
@@ -534,6 +534,7 @@ std::vector<DigitalSignature> Pkcs7Signature::getSignatures(const retdec::filefo
 	/* No signer would mean, we have pretty much nothing */
 	if (!signerInfo.has_value()) {
 		signatures.push_back(signature);
+		signature.isValid = false;
 		return signatures;
 	}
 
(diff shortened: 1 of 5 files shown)


src/fileformat/file_format/pe/authenticode/pkcs7_signature.cpp
```diff
@@ -556,18 +557,27 @@ std::vector<DigitalSignature> Pkcs7Signature::getSignatures(const retdec::filefo
 
 	signature.signer = signer;
 
+	// If there are warnings, then the signature is invalid
+	// We don't use the countersignature warnings here as
+	// previous implementation did when verifying
+	if (signature.warnings.empty()) {
```
metthal, line 563: Can this be simplified to `signature.isValid = signature.warnings.empty();`. It also reads nice as a sentence

> Signature is valid whenever there are no warnings.

metthal (changes requested):
General question, does `warnings` reflect also discrepancies like hash of the file does not match the hash in the signature, the signature of the hash is not the same as the one in authenticated attributes, ...? What about expiration? Previously we neglected the expiration dates as far as I remember and would consider the signature valid because if the hash matches and the signature is expired, we can't reliably tell that whether the signature is OK without verification in the whole chain and that's something we didn't want to do and we just stick with what we can verify statically. I would like to have a list of possible warnings and when can the warning occur before we proceed with this.

src/fileformat/file_format/pe/authenticode/pkcs7_signature.cpp
HoundThe: Yes, that's much better

HoundThe:
Expiration is not looked at, the only part regarding the certificate chain that is checked is the existence of the signing certificate.

Regarding the possible warnings, I'll split them into 2 parts - warnings regarding a signature and warnings regarding counter-signatures.

Signature warnings (PKCS7 structure):
- Couldn't parse the Pkcs7 signature"- Situation where the openssl library fails to parse the signature
- Wrong version
- Invalid number of DigestAlgorithmIdentifiers
- No contentInfo (structure where the actual signature digest is stored)
- Wrong contentInfo type
- Missing signature digest value
- Signature digest doesn't match the file digest
- No signerInfo
- No signer certificate
- Wrong signerInfo version
- Mismatching signerInfo and signature digest algorithms
- No encrypted digest
- No message digest
- Missing SpcSpOpusInfo structure
- Failure to verify the signature (decrypting encrypted digest and matching the digest - done by openssl `PKCS7_signatureVerify`)

CounterSignature warnings:

- Couldn't parse counter-signature.
- Missing signing certificate
- Missing content type (which again contains the digest)
- Failed to verify the counter-signature - calculating hash and comparing with the decrypted encrypted hash
- Unknown digest algorithm
- Failed to decrypt the digest
- Missing digest
- Failed to verify the signature with counter-signature - when the hash in the countersignature doesn't match with the hash of the countersigned signature.

metthal (approved):

HoundThe:
I have also removed version != 1 warning because there are samples with different versions that Windows accepts and it's not reflected by the old Authenticode specification.

PeterMatula:
lets run TC tests

PR #999, opened 2021-08-13
Switch to using Python3 module to detect a python

catap:
`PythonInterp` is old and deprecated module which may gave up on a system with multiple python installation like 2.7 and 3.x where `/usr/bin/python` is 2.7.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,7 +37,8 @@ include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/utils.cmake)
 include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/options.cmake)
 
 # RetDec, and some dependencies (e.g. LLVM, Keystone), require Python 3.
-find_package(PythonInterp 3.4 REQUIRED)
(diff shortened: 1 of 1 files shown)


catap:
Similar hack is required for macOS 10.15 with macports: https://github.com/macports/macports-ports/pull/11855

xkubov:
Lets run TC tests.

PeterMatula:
Thanks @catap, looks like a good change. It also looks that all of our CI configs don't like it, but we are investigating and hopefully will make them behave.

catap:
@PeterMatula it one of patches that is required to macport's.

xkubov:
So Linux and windows build should be good now - after rebasing onto current master all tests should pass. We must wait, however, for the update of the mac TeamCity agent before merging. The change requires a newer CMake version than we currently have installed and we have some problems with the mac agent at the moment. I will merge this after it is sorted out.

xkubov:
I have checked the TC builds and all of them are passing now (excluding Doxygen build). Therefore, I am merging this, thanks @catap.

PR #1006, opened 2021-08-27
Fixed discrepancies in icon hash between YARA and retdec-fileinfo

ladislav-zezula:


PeterMatula:
let's run TC tests

PeterMatula (approved):

PeterMatula:
The original test failed because this branch was not merged with `master`. When merged, they pass - except Doxygen, which is, in turn, failing in `master`.  For an unknown reason, tests are not published from TC to GitHub, so they are still red, even though they are green in TC -> merging anyway.

PR #1014, opened 2021-09-05
Fixed double using of CMAKE_INSTALL_PREFIX

catap:
`SUPPORT_TARGET_DIR` is defined at `set(SUPPORT_TARGET_DIR "${RETDEC_INSTALL_SUPPORT_DIR_ABS}")` when `RETDEC_INSTALL_SUPPORT_DIR_ABS` is defined as `set(RETDEC_INSTALL_SUPPORT_DIR_ABS  "${CMAKE_INSTALL_PREFIX}/${RETDEC_INSTALL_SUPPORT_DIR}")`.

If you put some real path inside `CMAKE_INSTALL_PREFIX`, for example `/path/to/somewhere`, it is installed all ords into `/path/to/somewhere/.../path/to/somewhere/ordinals/`.


xkubov:
Hi, thanks for the PR. I wasn't able, however, to reproduce your issue locally. There is no difference in installation paths on my system when I use `-DCMAKE_INSTLAL_PREIX=install` vs `-DCMAKE_INSTALL_PREFIX=$(realpath install)`.

What `cmake` are you using? In the code there should be, indeed, `RETDEC_INSTALL_SUPPORT_DIR_ABS` used and it should contain a correct absolute path all the time. The `*_ABS` variables are used through the project and we must ensure that they contain value as their name promise -> real path to some installation (sub)directory.

I looked into the definition and there is a variable `CMAKE_INSTALL_DATADIR` used that I think was expected to have a relative path `"share"` set all the time. The variable is set by CMake and is used to eliminate using `share` explicitly. Can you check that the variable has an absolute path set on your system? Can you also check `DATAROOTDIR ` variable?

catap:
@xkubov I've encountered this issue when I've created a port for MacPorts.

To build it on the right path and never leave sandbox I've hacked cmake related files like this:
```
post-patch {
    reinplace "s|\${CMAKE_INSTALL_PREFIX}|${destroot}\${CMAKE_INSTALL_PREFIX}|" ${worksrcpath}/CMakeLists.txt
    reinplace "s|\${CMAKE_INSTALL_PREFIX}|${destroot}\${CMAKE_INSTALL_PREFIX}|" ${worksrcpath}/support/CMakeLists.txt
    reinplace "s|DESTINATION \${SUPPORT_TARGET_DIR}/|DESTINATION \${RETDEC_INSTALL_SUPPORT_DIR}|" ${worksrcpath}/support/CMakeLists.txt
}
```

This PR is related to the last line, but I made introduce some site effects by the first two ;)

PR #1017, opened 2021-09-16
Modified dotNET's Yara rule

Dadda97:
_mscoree.dll_ is the only library needed for a .NET application and only one function is required to start the app.

Plus, the use of _??_ wildcard may slow down the scanning.


PeterMatula:
@ladislav-zezula what do you think?

ladislav-zezula:
At the moment, we wait for merging of an exact function, `pe.is_dotnet()` (merge request [here](https://github.com/VirusTotal/yara/pull/1568)). After this is done, we will replace checks for .NET with `pe.is_dotnet()` function.

PR #1027, opened 2021-09-24
Integrate new authenticode parser

HoundThe:
This PR replaces the current C++ Authenticode parser with https://github.com/avast/authenticode-parser
There are 2 main differences in the output: 

1. Only the first verification is reported. 
2. Only OnelineSubject/issuer is used and is just called as subject/issuer.

The new parser is also able to extract more nested signatures as it treats nested signatures correctly as a Set. But because the library doesn't have access to the PE file, I've moved the file digest calculation into the `pe_format.cpp`.


HoundThe:
I feel like Teamcity is not using the branch in regression-tests :thinking:

Issue #1028, opened 2021-09-26
ImageLoader::Save doesn't save Rich Header and section data


ladislav-zezula:
When `ImageLoader::Save()` is called, it produces skeleton of the PE file, containing mere PE headers. All section data are filled with zeros.

I'll prepare pull request for this.

PR #1029, opened 2021-09-26
Fixed ImageLoader::Save()

ladislav-zezula:
Fixed issue https://github.com/avast/retdec/issues/1028

### DIFF
--- a/src/pelib/ImageLoader.cpp
+++ b/src/pelib/ImageLoader.cpp
@@ -1053,9 +1053,8 @@ int PeLib::ImageLoader::Load(
 int PeLib::ImageLoader::Save(
 	std::ostream & fs,
 	std::streamoff fileOffset,
-	bool saveHeadersOnly)
+	FileFlags saveFlags)
 {
-	std::streamoff fileSize;
 	int fileError;
 
 	// Check and capture DOS header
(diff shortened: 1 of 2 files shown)


src/pelib/ImageLoader.cpp
```diff
@@ -1075,29 +1074,14 @@ int PeLib::ImageLoader::Save(
 		return fileError;
 
 	// Write zeros to the rest of the file, up to size of image
-	if(saveHeadersOnly == false)
+	if((saveFlags & FileFlags::HeadersOnly) == 0)
```
PeterMatula, line 1077: Is this equivalent to?
```cpp
if(!(saveFlags & FileFlags::HeadersOnly))
```
Wouldn't it be a little bit better with `!`?

src/pelib/ImageLoader.cpp
```diff
@@ -1761,109 +1757,13 @@ int PeLib::ImageLoader::saveNtHeaders(
 {
 	// Calculate the size of the optional header. Any version of PE file,
 	// 32 or 64-bit, must have this field set to a correct value.
-	std::uint32_t sizeOfOptionalHeader = getFieldOffset(PELIB_MEMBER_TYPE::OPTHDR_sizeof_fixed) + optionalHeader.NumberOfRvaAndSizes * sizeof(PELIB_IMAGE_DATA_DIRECTORY);
```
PeterMatula, line 1764: Here, a big chunk of code was removed. What was it doing and why don't we need it anymore?

src/pelib/ImageLoader.cpp
ladislav-zezula: Solved in https://github.com/avast/retdec/pull/1029/commits/228bacec87f99312653a05024ff4c6102f591ad9

src/pelib/ImageLoader.cpp
ladislav-zezula: This is because previous version reconstructed NT headers and then saved it to the file. This version saves the NT headers directly from the PE pages as-is.

PeterMatula:
lets run TC tests

ladislav-zezula:
let's run TC tests

PR #1030, opened 2021-09-26
Check if data is not empty in dotnet integer decoding functions

HoundThe:
Fix potential OOB read if passed data, to Signed/Unsigned dotnet decode function, is empty.


ladislav-zezula:
All asserts are gone and the change LGTM. Thank you for the fix! 👍

HoundThe:
Oh thank you for that, that was a mistake.

PeterMatula (approved):

PR #1036, opened 2021-10-26
Check for ELF damage

HoundThe:
Addition of very simple damage check - looking if the LOAD segments offset + filesize fits into the file.

### DIFF
--- a/src/fileformat/file_format/elf/elf_format.cpp
+++ b/src/fileformat/file_format/elf/elf_format.cpp
@@ -2062,6 +2062,25 @@ void ElfFormat::loadSections()
 	}
 }
 
+void ElfFormat::checkSegmentLoadable(const segment* seg)
+{
+	if (!seg)
+		return;
+
+	auto filesize = seg->get_file_size();
+	auto fileoffset = seg->get_offset();
+	int segtype = seg->get_type();
+
+	// Check if LOAD segments are actually in the file
+	if (segtype == PT_LOAD && getFileLength() < fileoffset + filesize)
+	{
+		_ldrErrInfo.isLoadableAnyway = false;
+		_ldrErrInfo.loaderErrorCode = ElfLoaderError::LDR_ERROR_SEGMENT_OUT_OF_FILE;
+		_ldrErrInfo.loaderError = "LOAD Segment data is not within file bounds";
+		_ldrErrInfo.loaderErrorUserFriendly = "LOAD Segment data is not within file bounds";
+	}
+}
+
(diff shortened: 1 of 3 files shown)


src/fileformat/file_format/elf/elf_format.cpp
```diff
@@ -2088,6 +2088,18 @@ void ElfFormat::loadSegments()
...
+		auto filesize = fSeg->getSizeInFile();
+		auto fileoffset = fSeg->getOffset();
+		int segtype = seg->get_type();
+
+		// Check if LOAD segments are actually in the file
+		if (segtype == PT_LOAD && getFileLength() < fileoffset + filesize)
+		{
+			_ldrErrInfo.loaderErrorCode = ElfLoaderError::LDR_ERROR_SEGMENT_OUT_OF_FILE;
+			_ldrErrInfo.loaderError = "Segment data is not within file bounds";
+			_ldrErrInfo.loaderErrorUserFriendly = "Segment data is not within file bounds";
```
PeterMatula, line 2101: You are not setting `isLoadableAnyway` here, and that thing is not set at `LoaderErrorInfo` constructor either. Could you please:
1. explicitly set it here, so we know what it is in this case.
2. init it in `LoaderErrorInfo` - probably to `false` to be more safe? It would be a little strange to have it `false` (by default) when `loaderErrorCode == 0` i.e. file is OK. But then when there is no error, consumers should not use these other values anyway and I think setting it to `False` (i.e. not loadable) is a more defensive when there actually is some problem - like in this ELF case.
   Moreover, feel free to refactor that `LoaderErrorInfo` code and use modern C++. I like this
   ```cpp
   std::uint32_t loaderErrorCode = 0;
   ```
   more than this:
   ```cpp
   LoaderErrorInfo() : loaderErrorCode(0)
   ```

src/fileformat/file_format/elf/elf_format.cpp
```diff
@@ -2088,6 +2088,18 @@ void ElfFormat::loadSegments()
 		fSeg->setElfAlign(seg->get_align());
 		fSeg->load(this);
 		segments.push_back(fSeg);
+
```
PeterMatula, line 2091: Maybe move this whole block to a separate function? Something like `checkSegment()`, `checkSegmentLoadable()`, etc.

PeterMatula (changes requested):
Could you please also add one regression test for this?

PeterMatula (approved):

Issue #1040, opened 2021-11-07
OpenSSL 3 support


catap:
An attempt to build `retdec` with local root f49ecb6ff6b8b220f3f2fa4b6b6354eeb59921e8 with OpenSSL 3 is failed:
```
:info:build Undefined symbols for architecture x86_64:
:info:build   "_EVP_MD_size", referenced from:
:info:build       _parse_authenticode in libauthenticode.a(authenticode.c.o)
:info:build       _pkcs9_countersig_new in libauthenticode.a(countersignature.c.o)
:info:build       _ms_countersig_new in libauthenticode.a(countersignature.c.o)
:info:build   "_EVP_PKEY_base_id", referenced from:
:info:build       _certificate_new in libauthenticode.a(certificate.c.o)
:info:build   "_TS_VERIFY_CTS_set_certs", referenced from:
:info:build       _ms_countersig_new in libauthenticode.a(countersignature.c.o)
:info:build ld: symbol(s) not found for architecture x86_64
```

PR #1041, opened 2021-11-07
Update API for OpenSSL 3.0

catap:
Here a minor changes in OpenSSL API.

Fixes: https://github.com/avast/retdec/issues/1040

### DIFF
--- a/deps/authenticode-parser/src/authenticode.c
+++ b/deps/authenticode-parser/src/authenticode.c
@@ -581,7 +581,11 @@ AuthenticodeArray* parse_authenticode(const uint8_t* pe_data, long pe_len)
             continue;
         }
 
+#if OPENSSL_VERSION_NUMBER >= 0x3000000fL
+        int mdlen = EVP_MD_get_size(md);
+#else
         int mdlen = EVP_MD_size(md);
+#endif
         sig->file_digest.len = mdlen;
         sig->file_digest.data = (uint8_t*)malloc(mdlen);
         if (!sig->file_digest.data)
--- a/deps/authenticode-parser/src/certificate.c
+++ b/deps/authenticode-parser/src/certificate.c
@@ -287,7 +287,11 @@ Certificate* certificate_new(X509* x509)
     EVP_PKEY* pkey = X509_get0_pubkey(x509);
     if (pkey) {
         result->key = pubkey_to_pem(pkey);
+#if OPENSSL_VERSION_NUMBER >= 0x3000000fL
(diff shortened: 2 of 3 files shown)


s3rvac:
@PeterMatula Do we want to use OpenSSL 3.0 in `authenticode_parser` when (1) the rest of RetDec uses OpenSSL 1.x and (2) many systems only have OpenSSL 1.x?

catap:
@s3rvac I feel that the right way is requires the same `OpenSSL` version everywhere. Right now it seems quite a mix. But I will wait of @PeterMatula response.

Anyway all another part of `retdec` is built fine with OpenSSL 3.

catap:
From another hand I may easy add a few `ifdef` to make the code universal.

catap:
Here it is. Right now it just introduced a compatibility with OpenSSL 3 without broken anything :)

catap:
@s3rvac I feel that I've overstep all your concern, am I wrong?

s3rvac:
Yes, thank you :slightly_smiling_face:. I will leave the review to @PeterMatula.

catap:
@s3rvac / @PeterMatula I do also have a patch for version 4.0: https://github.com/catap/retdec/tree/openssl3-v4.0

Do you need as PR?

PeterMatula (approved):
Looks good, I will trigger tests and if they pass I will merge.

Btw, these sources in `deps` are in fact from here [avast/authenticode-parser](https://github.com/avast/authenticode-parser), but because we wanted to reduce the number of external deps, we decided to _"inline"_ this simple repo in RetDec directly. But not to worry, @HoundThe can you please integrate these changes back to [avast/authenticode-parser](https://github.com/avast/authenticode-parser)?

And a second task, @HoundThe can you look into the question of _"different OpenSSL versions"_ opened here? I would like to know, where (what other parts) are we using OpenSSL, which versions, and if they can all deal with OpenSSL 3.0.

catap:
@PeterMatula thanks!

Anyway, I'd love to share with you that v4.0 can't be build on system which has installed OpenSSL 3.0. Yes, it ships OpenSSL but system one is also used to build it. When your system has 1.x it is working, but with 3.0 => the hell is here.

The way to fix it is trivial: https://github.com/catap/retdec/commit/6aa99feff5ea9290a0983633d26f8ad85fdae5ba

But it still, this patch need to be applied.

v4.0 also can't be build on Apple M1 because of embeded OpenSSL version which is too old.

And here I'd love to ask: maybe it is a good idea to release a new version?

PR #1051, opened 2022-01-12
Fixed false positive in the detection of PyInstaller 3.x

ladislav-zezula:
* YARA rule for PyInstaller 3.x no longer detects DLLs like `602a21507939a71fe3ac045b81498a0f6a6dac68da45491a10950019fcd154a8`
* YARA rule for PyInstaller 3.x now also detects EXEs created with newer version of PyInstaller


metthal (approved):
LGTM

PR #1052, opened 2022-01-14
Fix: Manually-specified variables were not used by the project.

xkubov:
The following variables were set in the CMakeLists.txt, however, they were not used by the LLVM project build:

LLVM_USE_CRT_DEBUG
LLVM_USE_CRT_RELEASE


xkubov:
Lets run TC tests

PR #1055, opened 2022-01-17
Use image loader when loading corrupted resources

metthal:
When the resource tree doesn't have expected depth, we use different
path to load resources which didn't use loader but calculated offset
from RVA on it's own resulting in underflows on integers and completely
bogus offsets. Even though offsets are still kinda bogus, they make
sense and are within the file itself after this change.

Example file: f39f1ad9f97766618826d47c2d32529425e773659f5577fea06a82e21f13b064


xkubov:
Lets run TC tests

xkubov (approved):

PR #1063, opened 2022-02-17
Lz include relocation into image load

ladislav-zezula:


ladislav-zezula:
let's run TC tests

s3rvac (approved):
Thank you for the fix. All tests are passing, so I am going to merge the PR, including the newly added regression test in https://github.com/avast/retdec-regression-tests/pull/115.

PR #1067, opened 2022-03-09
Updated authenticode parser to the newest version

metthal:


HoundThe (approved):
Thanks for the update! I've added some additional fixes due to failed test. I will merge after finishing the test suite.

PR #1073, opened 2022-04-18
Change procedure linkage table indirect calls to direct calls

richardlford:
When a function in an external dynamically-loaded library is
called, the call is made indirectly through the
procedure linkage table (PLT). The call into the PLT is
made with an ordinary call instruction, but in the PLT
there is a jump (not call) to the final destination.
In this situation retdec is not properly handling
the parameter passing. This change turns these indirect
calls into direct calls and deletes the PLT code.
This makes the result simpler and allows the parameters
to be properly passed.

### DIFF
--- a/include/retdec/bin2llvmir/optimizations/decoder/decoder.h
+++ b/include/retdec/bin2llvmir/optimizations/decoder/decoder.h
@@ -337,6 +337,46 @@ class Decoder : public llvm::ModulePass
 		bool _switchGenerated = false;
 
 		bool _somethingDecoded = false;
+
+	// Fixing Procedure Linkage Table (PLT) calls.
+	private:
+		// Top level function for fixing PLT calls.
+		//
+		// When a call is made to an external function, e.g. C++ operator new,
+		// the call is resolved by directing the call to an entry in the
+		// Procedure Linkage Table (PLT). For example, this call
+		//
+		// 12d1:	callq  1130 <_Znwm@plt>
+		//
+		// goes to this entry:
+		//
+		// Disassembly of section .plt.sec:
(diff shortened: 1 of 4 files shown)


PeterMatula:
I will, before the end of this week.

PeterMatula:
Thanks @richardlford, I think in general it looks good. Few regression tests are failing, but I reviewed the problems and it looks like it is nothing related to output quality regression. Most fails are due to some crash I will have to further investigate, and probably one or two are some minor details that were testing the layout before your changes.
I will try to fix the tests and let you know.

richardlford:
> Few regression tests are failing, but I reviewed the problems and it looks like it is nothing related to output quality regression. Most fails are due to some crash I will have to further investigate, and probably one or two are some minor details that were testing the layout before your changes.
> I will try to fix the tests and let you know.

@PeterMatula, sorry, I should have run the tests first. I am now setting up to run the regression tests myself.

richardlford:
@PeterMatula, I have investigated the crashes. In my code, after making the PLT calls direct I was deleting the PLT function. But that was causing problems related to virtual calls. When I don't delete the PLT function only 3 tests are failing and those are ones where my changes cause improvement, so the test will need to be adjusted. I expect to have a new commit early next week.

richardlford:
@PeterMatula, I did decide to accomplish my goal in the param_return pass rather than with these changes. Since I don't use any of these changes I'm closing this pull request. I've created a new pull request, #1085. I've also created a pull request in the retdec-regressions-tests repo, https://github.com/avast/retdec-regression-tests/pull/117#issue-1297600973, to account for the effect of these changes.

Issue #1045, opened 2021-12-02
retdec-decompiler fails on macOS 12


catap:
When I try to decompile any binary on macOS 12, it crashes:
```
libc++abi: terminating with uncaught exception of type std::runtime_error: failed to limit maximal memory to half of system RAM
0  retdec-decompiler        0x000000010aed070a llvm::sys::PrintStackTrace(llvm::raw_ostream&) + 58
1  retdec-decompiler        0x000000010aed0c59 PrintStackTraceSignalHandler(void*) + 25
2  retdec-decompiler        0x000000010aecdaa3 llvm::sys::RunSignalHandlers() + 115
3  retdec-decompiler        0x000000010aed69ac SignalHandler(int) + 204
4  libsystem_platform.dylib 0x00007ff81a5a6e2d _sigtramp + 29
5  libc++abi.dylib          0x00007ff81a550f4c
6  libsystem_c.dylib        0x00007ff81a4ddd10 abort + 123
7  libc++abi.dylib          0x00007ff81a54e0b2 abort_message + 241
8  libc++abi.dylib          0x00007ff81a53f1e5 demangling_terminate_handler() + 242
9  libobjc.A.dylib          0x00007ff81a43c511 _objc_terminate() + 104
10 libc++abi.dylib          0x00007ff81a54d4d7 std::__terminate(void (*)()) + 8
11 libc++abi.dylib          0x00007ff81a54fd55 __cxa_get_exception_ptr + 0
12 libc++abi.dylib          0x00007ff81a54fd1c __cxxabiv1::exception_cleanup_func(_Unwind_Reason_Code, _Unwind_Exception*) + 0
13 retdec-decompiler        0x000000010a3133ce limitMaximalMemoryIfRequested(retdec::config::Parameters const&) + 84
14 retdec-decompiler        0x000000010a31527b main + 554
15 dyld                     0x000000010fe994fe start + 462
16 dyld                     000000000000000000 start + 4028001488
17 retdec-decompiler        0x000000010a30b000 __dso_handle + 0

```

anyway, `--max-memory 0` works fine

xkubov:
Hi, this issue is somehow related to https://github.com/avast/retdec/issues/379. The issue in #379 is that setting a limit on macOS is not doing anything. In your case it is doing something - failing :D 

[...]

Issue #379, opened 2018-08-23
Memory limiting does not work on macOS


s3rvac:
The limiting of memory via `--max-memory` does not work properly on macOS. For example, if you run
```
$ retdec-fileinfo --max-memory=4096 FILE
```
the analysis (properly) fails on Linux and Windows, but finishes successfully on macOS. The reason is that on macOS, the following piece of code from `src/utils/memory.cpp` compiles, runs, returns 0 (success), but does actually not do anything:
```cpp
struct rlimit rl = {
    .rlim_cur = limit,        // Soft limit.
    .rlim_max = RLIM_INFINITY // Hard limit (ceiling for rlim_cur).
};
auto rc = setrlimit(RLIMIT_AS, &rl);
```
From what I was able to find, `setrlimit()` with `RLIMIT_AS` is broken on macOS and simply does not work.

What is the proper way of programmatically limiting virtual memory on macOS? Feel free to submit a [PR](https://github.com/avast-tl/retdec/pulls) that fixes the implementation of `limitSystemMemoryOnMacOS()` in `src/utils/memory.cpp`.

s3rvac:
In #1074, we have disabled the limiting of the maximum memory on macOS until there is a working way of doing so.

PR #1074, opened 2022-04-22
Never try to limit memory on macOS

catap:
We can't limit memory on macOS. Before macOS 12 limitSystemMemoryOnPOSIX() does not actually do anything on macOS. Anyway, it just succeed, since macOS 12 it returns error and retdec can't start.

To be honest Apple can control memmory limit via so-called the ledger() system call which is private. An old verison which was opened to OpenSource (from 10.9-10.10?) using setrlimit() but at some point setrlimit() was broken and not ledger(). Probably at macOS 12 the setrlimit() was completley broken.

Because we haven't got any other choose just return true which haven't change anything.

See: https://github.com/avast/retdec/issues/379
Fixes: https://github.com/avast/retdec/issues/1045

### DIFF
--- a/src/utils/memory.cpp
+++ b/src/utils/memory.cpp
@@ -27,7 +27,7 @@ namespace utils {
 
 namespace {
 
-#ifdef OS_POSIX
+#if defined(OS_POSIX) && !defined(OS_MACOS)
 
 /**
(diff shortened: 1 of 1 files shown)


s3rvac (approved):
Thank you for the update. I agree that we should not try to limit the maximum amount of memory on macOS until there is a working way of doing so.

PR #1081, opened 2022-06-08
Use multistage build for Dockerfile, reduces container size by ~1.5G

bagelbyte:
This commit essentially just removes the build dependencies/source from the final docker image that gets created, which reduces the size of the built image from ~3.2GB to ~1.6GB as of the current master branch


PeterMatula (approved):

PeterMatula:
Thanks a lot @bagelbyte :+1:

PR #1059, opened 2022-01-26
[Capstone2llvmir] Update to Capstone V5

owlxiao:
# [deps][capstone][CMakeLists]
## Add CAPSTONE_INSTALL opinion
Generate install target

# [Capstone2LlvmIr][PowerPC]
## Remove ppc_reg_cr_flags enum
In capstone V5, each bits of CR field already has native support

## Fix translateClrlwi method
In capstone V5, rlwinm is equivalent to clrlwi

## Add branch mnemonics incorporating conditions
        BLT*        BLE*        BEQ*        BGE*
        BGT*        BNL*        BNE*        BNG*
        BSO*        BNS*

For BUN* and BNU*, they are equivalent to BSO*/BNS*,
because they are read same CR field

## Add PPC_REG_ZERO
The reprsentation of r0 when treated as the constant 0.

## Fix translateCrSetClr method
In capstone V5, the crset/crclr operator will return detail field of CR,
            so we should judge CR field instead of CR

## Add isCrBitRegister()
judge whether it is a field of CR

## Fix Crand Unit Test

# [Capstone2LlvmIr][Arm]
## Fix ARM_INS_NOP Unit test
In capstone V5, nop is a hint instruction.

# [Capstone2LlvmIr][Arm64]
## Add following instructions support:
set flags:
    
    ARM64_INS_ADDS, ARN64_INS_ANDS,
    ARM64_INS_BICS, ARM_INS_SBCS  ,
    ARM_INS_SUBS

## Remove op.VESS
it overlaps with op.VAS

# [Capstone2LlvmIr][X86]
## Delete X86_INS_PCOMMIT
The PCOMMIT instruction has been deprecated.

## Remove following instructions

    X86_INS_VCVTPD2DQX, X86_INS_VCVTPD2PSX, X86_INS_VCVTTPD2DQX

each pseudo instruction of:

        X86_INS_CMPSS
        X86_INS_CMPSD
        X86_INS_CMPPS
        X86_INS_CMPPD
        X86_INS_VCMPSS
        X86_INS_VCMPSD
        X86_INS_VCMPPS
        X86_INS_VCMPPD

## Fix loadOpFloatingBinaryTop
when transalte "FXCH instruction, in the value of loadOpFloatingBinaryTop Function,
            "top" is equivalent to idx, which causes the value to be written to top twice when exchanging data.

## Fix translateFadd method
in capstone V5, FADDP = FADD. so we should judge it by read opcode

PR #1086, opened 2022-07-07
Upgrade to Capstone release 4.0.2 with patch

richardlford:
This is the latest released version of Capstone and the first Capstone release to support the endbr32 and endbr64 instructions. I was trying to add some new tests for RetDec, but although I can use an option to avoid producing the endbr instructions, the library has them so they end up in the produced binary. Thus if we want to add any new tests (which I do), we need to update to a newer Capstone.

### DIFF
--- a/cmake/deps.cmake
+++ b/cmake/deps.cmake
@@ -1,10 +1,11 @@
 
+# URL is for Capstone release 4.0.2.
 set(CAPSTONE_URL
-    "https://github.com/aquynh/capstone/archive/bc8a649b35188786754ea1b0bddd5cb48a039162.zip"
+    "https://github.com/capstone-engine/capstone/archive/1d230532840a37ac032c6ab80128238fc930c6c1.zip"
     CACHE STRING "URL of Capstone archive to use."
 )
 set(CAPSTONE_ARCHIVE_SHA256
-    "7d3075bce1f5622279c16a6f62fe8c548d4544bfc82292f6bf43907d0317fd10"
+    "659097fcda59ce927937f73dd87a4606de6e768b352045a077ed8d2165b7e935"
     CACHE STRING ""
 )
 
--- a/deps/capstone/patch.cmake
+++ b/deps/capstone/patch.cmake
@@ -11,9 +11,16 @@ string(REPLACE
     "${content}"
 )
 
-if("${new_content}" STREQUAL "${content}")
+string(REPLACE
+    "static void printS16ImmOperand(MCInst *MI, unsigned OpNo, SStream *O)\n{\n\tif (MCOperand_isImm(MCInst_getOperand(MI, OpNo))) {\n\t\tunsigned short Imm = (unsigned short)MCOperand_getImm(MCInst_getOperand(MI, OpNo));\n        if (Imm > HEX_THRESHOLD)\n            SStream_concat(O, \"0x%x\", Imm);\n        else\n            SStream_concat(O, \"%u\", Imm);\n"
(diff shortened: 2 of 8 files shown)


richardlford:
@PeterMatula, please review this change. This change causes around 50 test failures in the regression tests. I will be examining the changes to see if they are really regressions, or just differences, or possibly improvements.

richardlford:
@PeterMatula, as an alternative to doing a full upgrade to the 4.0.2 release, I experimented with forking capstone and just cherry-picking the endbr* support onto the version of capstone that RetDec is currently using (See https://github.com/richardlford/capstone/commit/a6ee24f01f9713a5ef61056c68eb05a79ffbe2c1). I then modified RetDec to use that capstone and also added code to treat the endbr* instructions as NOPs (i.e. producing no LLVM). The result passes all of the RetDec regression tests. If you don't want to upgrade to the full 4.0.2 release of capstone you could consider having your own fork of capstone and apply the endbr* support change as I did. Of course, it seems preferable to try to keep up with capstone otherwise RetDec is likely to have continuing problems with newer binaries.

Thanks for your help.

richardlford:
@PeterMatula, I have pushed another commit that patches a bug in capstone 4.0.2. In addition to fixing the bug, the patch also now always prints the 16-bit immediates in decimal (this is the behavior in a later version of capstone that has this bug fixed0. With this patch, all of the RetDec regression tests passed except one that was expecting the immediates to be printed in hex. I have submitted a separate PR (https://github.com/avast/retdec-regression-tests/pull/118) to fix that test, so you will want to merge that PR after this one.

Could you please review it?

Thanks

PeterMatula:
Thanks for this, I will definitely look at it, as I planned to update Capstone for a long time but didn't get around to it myself. I'm not against updating Capstone even if it causes some tests to fail - we will either fix them if possible or disable them. It is not good to remain in a state of the old Capstone version just for the sake of tests.

Some time ago I also experimented with Capstone v5 and there is also #1059 PR, but so far I didn't manage to get the tests to some reasonable state with the new Capstone. I would prefer to use the latest Capstone possible, but I'm not sure what is the state of Capstone project, if the `next` will even be merged to `master`, what version are Capstone users supposed to use, and if it is worth the effort to move to v5 as it requires even more changes than `4.0.2`. Do you have any opinion/knowledge on this?

richardlford:
> Some time ago I also experimented with Capstone v5 and there is also #1059 PR, but so far I didn't manage to get the tests to some reasonable state with the new Capstone. I would prefer to use the latest Capstone possible, but I'm not sure what is the state of Capstone project, if the `next` will even be merged to `master`, what version are Capstone users supposed to use, and if it is worth the effort to move to v5 as it requires even more changes than `4.0.2`. Do you have any opinion/knowledge on this?

@PeterMatula, the 4.0.2 version is the latest `released` version (i.e. with a github release), though there are tagged v5 release candidates. I tried various newer versions but was getting compile errors building RetDec with them. I think moving to 4.0.2 would be a good interim step toward newer versions. I had the impression that the `next` branch was going to be the permanent branch for development, perhaps to avoid the name `master` because of it social implications. 

And the 4.0.2 version supports the endbr32 and endbr64 instructions. This is essential to handle some newer binaries. So I'd appreciate it if you would ahead and merge this. 

In studying Capstone's use of the llvm-mc infrastructure, it occurred to me that RetDec, with some effort, might be able to eliminate the use of Capstone and just use LLVM's decoder directly, i.e. it McInsts. But that would require further study. But LLVM is very active and more likely to have up-to-date instructions.

PeterMatula (approved):

PeterMatula:
@richardlford thanks a lot for this, you did what I was putting off for a long time. Especially that Capstone patch is not an easy thing to figure out.

I did one more change in 2771f85 as this code was causing problems once the issue in Capstone was fixed.

PeterMatula:
Damn, by accident I did _Rebase and merge_ which was not the best option for this PR. Hope you don't mind. I can try to revert it and do it again, but I'm afraid that would only further complicate things.

richardlford:
> Damn, by accident I did _Rebase and merge_ which was not the best option for this PR. Hope you don't mind. I can try to revert it and do it again, but I'm afraid that would only further complicate things.

No problem. You can leave it as it is.

PR #1088, opened 2022-07-18
Handle Procedure Linkage calls for 32bit x86 from gcc

richardlford:
This case is for x86 32 bit compiled with GCC. Its PLT entries are in
sections .plt.sec or .plt.got. An entry is of the form:

jmp *offset(%ebx)

When this code is encountered register %ebx has been loaded with the
address of the start of the Global Offset Table (.got) section.
This change handles that case.


PeterMatula (approved):

PeterMatula:
Again, thanks @richardlford for the change, and double for the tests.

PR #1090, opened 2022-07-22
Add ability to process PNG icons for perceptual hash calculation

HoundThe:
I've integrated https://github.com/nothings/stb `stb_image.h` to add loading of PNG icons for perceptual hash calculation. I've flipped the PNG image upside down, as the existing DIB format is stored upside down, and I wanted to align with the current implementation.

I've handcrafted some test samples to the appropriate repository with mention of this PR. I took a sample with PNG Icon, extracted it in a different resolution, converted it to DIB format, and put it inside another file. Then I test if the perceptual hash of both file match.

### DIFF
--- a/src/fileformat/types/resource_table/bitmap_image.cpp
+++ b/src/fileformat/types/resource_table/bitmap_image.cpp
@@ -8,6 +8,7 @@
 
 #include "retdec/fileformat/types/resource_table/bitmap_image.h"
 #include "retdec/utils/conversion.h"
+#include "retdec/utils/scope_exit.h"
 #include "retdec/utils/system.h"
 #include <stb/stb_image.h>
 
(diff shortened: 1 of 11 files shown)


src/fileformat/types/resource_table/bitmap_image.cpp
```diff
@@ -47,6 +48,30 @@ const std::vector<std::vector<struct BitmapPixel>> &BitmapImage::getImage() cons
...
+		std::vector<BitmapPixel> row;
+		for (int j = 0; j < x; j++)
+		{
+			int offset = (i * x + j) * 4;
+			row.emplace_back(data[offset], data[offset + 1], data[offset + 2], data[offset + 3]);
+		}
+		this->image.push_back(row);
+	}
+
+	stbi_image_free(data);
```
metthal, line 70: We have a `ScopeExitGuard` in `utils/scope_exit.h`. Check it out. I would utilize it here as it should provide more safety when it comes to exceptions etc.

metthal (changes requested):

src/fileformat/types/resource_table/bitmap_image.cpp
HoundThe: Thanks, TIL. I am not 100% if I used it correctly, but I've tried to utilize it.

PR #1092, opened 2022-08-04
Add prototypes for dynamically-linked functions without headers

richardlford:
When the program involves dynamically-linked functions like _Znwj
(operator new) that return a pointer, it is necessary to have
prototypes for them, since otherwise they will be implicitly deduced
to return "int" which cannnot be dereferenced.

Previously RetDec was emitting comments telling which functions were
dynamically linked. This change moves them up before the functions are
emitted and instead emits prototypes for the functions. However,
RetDec also inserts includes of headers for functions with known
headers. So this change does not emit prototypes for functions with headers as that
would be redundant.  As a result, some dynamically-linked functions
that used to show in the comments no longer appear as the included
header will declare them.

The section header comment for dynamically-linked functions is only
produced if some prototypes are written for dynamically-linked
functions.

A related PR (https://github.com/avast/retdec-regression-tests/pull/121) 
adds tests as well as changes needed for existing tests.

### DIFF
--- a/include/retdec/llvmir2hll/hll/hll_writer.h
+++ b/include/retdec/llvmir2hll/hll/hll_writer.h
@@ -91,6 +91,7 @@ protected:
 
 	virtual bool emitFunctionPrototypesHeader();
 	virtual bool emitFunctionPrototypes();
+	virtual bool emitFunctionPrototypes(const FuncSet &funcs);
 
 	virtual bool emitFunctionsHeader();
 	virtual bool emitFunctions();
@@ -99,7 +100,6 @@ protected:
 	virtual bool emitStaticallyLinkedFunctionsHeader();
 	virtual bool emitStaticallyLinkedFunctions();
 
-	virtual bool emitDynamicallyLinkedFunctionsHeader();
 	virtual bool emitDynamicallyLinkedFunctions();
 
 	virtual bool emitSyscallFunctionsHeader();
--- a/src/llvmir2hll/hll/hll_writer.cpp
+++ b/src/llvmir2hll/hll/hll_writer.cpp
@@ -33,6 +33,7 @@
 #include "retdec/llvmir2hll/llvm/llvm_support.h"
 #include "retdec/llvmir2hll/support/debug.h"
 #include "retdec/llvmir2hll/support/global_vars_sorter.h"
+#include "retdec/llvmir2hll/support/headers_for_declared_funcs.h"
 #include "retdec/llvmir2hll/support/smart_ptr.h"
 #include "retdec/llvmir2hll/utils/ir.h"
 #include "retdec/llvmir2hll/utils/string.h"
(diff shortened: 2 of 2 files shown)


richardlford:
@PeterMatula, could you please review this change? This continues my effort to get dynamically-linked functions to decompile correctly.

PeterMatula:
Thanks @richardlford for another improvement. In general I like it, however did you notice a small regression in [this](https://github.com/avast/retdec-regression-tests/pull/121/files#diff-6b22faa122787a53b4957bd39c9255b438a9ad0046763162333c0297d1791941L13) test? Looks like before, we applied type info to functions in `Dynamically Linked Functions` section, but now we don't do so for the ones in `Dynamically Linked Functions Without Header`. Were you aware of this? Wouldn't it be better if we tried to apply known type info to these functions, or would it potentially backfire in some cases? I can try to identify the place where the info is applied if needed.

richardlford:
> In general I like it, however, did you notice a small regression in [this](https://github.com/avast/retdec-regression-tests/pull/121/files#diff-6b22faa122787a53b4957bd39c9255b438a9ad0046763162333c0297d1791941L13) test? Looks like before, we applied type info to functions in `Dynamically Linked Functions` section, but now we don't do so for the ones in `Dynamically Linked Functions Without Header`. Were you aware of this? Wouldn't it be better if we tried to apply known type info to these functions, or would it potentially backfire in some cases? I can try to identify the place where the info is applied if needed.

I have investigated this further. Here are some findings:

- For the test that you referenced, the link-time-information in support/types/windrivers.json give the header for the dynamically-linked functions (e.g. KeStallExecutionProcessor is declared in minwin/ntosp.h), but the llvmir2hll pass does not use that information but uses the information from retdec::llvmir2hll::Semantics class and its subclasses. The WinAPISemantics class has information on some Windows headers but does not have it for those mentioned in windrivers.json. It seems like it would be useful if the information on headers were centralized. I could add the information about headers from windrivers.json and add it to the WinAPISemantics class. Do you want me to do that? If I did all of the dynamically-linked functions in this test would have an associated header and thus would disappear.
- The former comments about dynamically-linked functions (to be continued in a further edit)

richardlford:
@PeterMatula, ping?

PeterMatula:
Sorry about the delay @richardlford, I know about it, I just didn't get around to looking into it more so I can answer. I have a look tomorrow and let you know.

PeterMatula:
Sorry for taking so long @richardlford, and thanks for the improvement.

I looked into it again and realized I misunderstood what exactly was going on the first time. I thought there is a regression on an LLVM IR level, and that we are possibly losing correct data type information on function parameters. But that is not the case, all the info is there, we just generate it slightly differently. The output could be further improved with the info we have by either omitting all the linked functions we know via LTI, or by using the argument names from the same source. But this is not the goal of your PR, so let's not complicate it. You are right about this:
> It seems like it would be useful if the information on headers were centralized.

Or even better, it would be best to use external knowledge like LTI information once somewhere in the beginning and embed it all into LLVM IR either directly or via some well-documented metadata, so that we have a fully self-contained LLVM IR and further passes do not have to access specialized resources. But like I said, this is just an idea for further improvements, it does not block your work.

PR #1107, opened 2022-09-08
Add printing of analysis time to retdec-fileinfo output

metthal:
This adds printing of the time of an analysis into both plain and JSON output of `retdec-fileinfo`. It only does it when `--analysis-time` command line options is used if anyone ever relied on the fact that the output of `retdec-fileinfo` is always the same.


HoundThe (approved):
Looks good! The Mac test failed, but it seems to come from the decompiler, so I've restarted it to see if it persists.

PR #1115, opened 2022-09-20
Simplify removing range

neverwin:
The logic remains the same.


PeterMatula:
Thanks for the improvement.

PR #1120, opened 2022-11-02
Added sanity check for page index when loading pages from broken samples

metthal:
There are certain samples where page index might go beyond available pages when trying to load them which will be prevented with this patch.


ladislav-zezula:
I added proper handling. Also, retdec-fileinfo now reports loader error properly.

metthal:
LGTM

PR #1123, opened 2022-11-13
Allow to configure `yara`'s make tool

catap:
This is another approach to remove hardocoded call of `make` to build `yara`. This approach uses the same `make` by default but introduces `YARA_MAKE_PROGRAM` option which can be used to specify a path to `gmake` for example.

Thus, when CMake produces a set of makefile, it reuses `CMAKE_MAKE_PROGRAM` to respect a global `-jXXX` option.


PeterMatula:
Thanks, although the tests fail at the moment, I merged it - it looks fine, and the tests are broken - we are working on it.

catap:
@PeterMatula I've used this patch at MacPorts for a while => it is fine :)

PR #1136, opened 2023-02-28
Do not transform TLSH and provide it as-is (uppercase)

metthal:
These kind of hashes are always provided as upper-case (for example on VirusTotal). We are the only one providing them in lower-case and we are doing it deliberately. I suggest not to do it and provide them as-is and that is uppercase.

### DIFF
--- a/src/fileformat/file_format/elf/elf_format.cpp
+++ b/src/fileformat/file_format/elf/elf_format.cpp
@@ -1902,7 +1902,7 @@ void ElfFormat::loadTelfhash()
 
 		tlsh.final();
 		const int show_version = 1; /* this prepends the hash with 'T' + number of the version */
-		telfhash = toLower(tlsh.getHash(show_version));
+		telfhash = tlsh.getHash(show_version);
 	}
 }
 
--- a/src/fileformat/types/import_table/elf_import_table.cpp
+++ b/src/fileformat/types/import_table/elf_import_table.cpp
@@ -49,7 +49,7 @@ void ElfImportTable::computeHashes()
 		tlsh.final();
 		/* this prepends the hash with 'T' + number of the version */
 		const int show_version = 1;
-		impHashTlsh = toLower(tlsh.getHash(show_version));
+		impHashTlsh = tlsh.getHash(show_version);
 
 		impHashCrc32 = getCrc32(data, impHashString.size());
 		impHashMd5 = getMd5(data, impHashString.size());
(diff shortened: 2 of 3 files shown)


PeterMatula:
An associated https://github.com/avast/retdec-regression-tests/pull/124 PR which should fix the tests.

PeterMatula (commented):
The code is fine, but...

We probably transformed it to lower case so that it is consistent with all the other hashes we produce. There is no functional problem with this change, but you break this consistence. See [this[(https://github.com/avast/retdec-regression-tests/pull/124/files#diff-edb3f0505d285d2a2265e6dd1af8488cb5b6cb191ce1e15b17a6c375b26b83aeL205).

I see several options:
* We leave it as-is in this PR - `telfhash` & `importTable.tlsh` will be in different case that other hashes.
* We transform `importTable.tlsh` to lower hash so that it is consistent with other `importTable.*` hashes. But `telfhash` stays upper case.
* We transform all the hashes to upper case.
* We don't accept this PR and all the hashes remain lower case.

@metthal we can discuss on slack or zoom.

metthal:
I am aware that we are breaking consistency but the other option is that we are breaking compatibility with other users of TLSH. On top of that, what we are right now generating is not valid TLSH hash because we are lowercasing even the version prefix `T1` - which is expected to be that way - see [their code](https://github.com/trendmicro/tlsh/blob/master/src/tlsh_impl.cpp#L796). Even VirusTotal or others provide SHA256 in lower case while TLSH is provided as-is, in uppercase.

> We transform importTable.tlsh to lower hash so that it is consistent with other importTable.* hashes. But telfhash stays upper case.

I wouldn't want this.

> We transform all the hashes to upper case.

Wouldn't want this either.

> We don't accept this PR and all the hashes remain lower case.

YARA provides them in uppercase and I can certainly approach this problem from the other side and try to push YARA upstream to change it to lower case but I don't really know how to sell this. The only complaint here is consistency with other hashes, but I really wonder whether that's worth considering that you won't for example be able to copy TLSH from VirusTotal as-is and put it into YARA rule (that's essentially complaint here internally by the researchers that the output of RetDec cannot be directly copied to YARA). We can also change just the internal YARA to lowercase but that would break compatibility with upstream. Considering all the other options before, the conclusion I came to was to eliminate the black sheep - RetDec, which is the only one doing this differently than others.

PeterMatula (approved):
@metthal Fair enough, I didn't know the state in other projects, and if it is like that, I see no reason to not generate uppercase.

Issue #1148, opened 2023-04-22
Error "Missing type for register number"


Nitr0-G:
Hello everyone

I'm trying to lift UnPackMe_ACProtect Pro 2.1.0.exe after several phases I get an error with further output
```
Running phase: Unpacking ( 0.01s )
No matching plugins found for 'Perplex PE Protector'
No matching plugins found for 'Borland'
No matching plugins found for 'ASProtect 1.2'
Running phase: Initialization ( 0.05s )
Running phase: Providers initialization ( 0.06s )
Running phase: Input binary to LLVM IR decoding ( 0.53s )
[capstone2llvmir]: Missing type for register number: 242 ;<- Crash
```
Pe file(UnpackMe from tuts4you collection 2011)
[UnPackMe_ACProtect Pro 2.1.0.zip](https://github.com/avast/retdec/files/11299901/UnPackMe_ACProtect.Pro.2.1.0.zip)

What could be the reason for this behavior? Arguments for start:
 
**retdec-decompiler.exe --arch=x86 "UnPackMe_ACProtect Pro 2.1.0.exe"**

Nitr0-G:
Will anyone answer?

Nitr0-G:
I found what the problem is, this instruction is encoded due to static analysis of `bndldx bnd0, ss:[ebp+0x5EA63807]`, there is no such instruction. My idea is to convert retdec to a dynamic lifter so that there are no such situations

Nitr0-G:
I have studied the work of the project, everything related to the specific data of the section/sections is initialized here

```
	initTranslator();
	initDryRunCsInstruction();
	initEnvironment();
	initRanges();
	initJumpTargets();
```

Next, go to the `decode()` function(`decoder.cpp`), in the While loop in this function, the main processing of instructions occurs. To go from instruction to instruction and process it, we go to `Decoder::decodeJumpTarget(const JumpTarget& jt)`, where the `start` object of the `Address` class is initialized, on line 217 `decoder.cpp` we get the first data in the `bytes` object of the `ByteData` class, we get them from the `getRawSegmentData(start)` method

Nitr0-G:
I have figured out the problem in more detail and have the following information:
![eecd61a09455d0ef3016e34430e2b6a5](https://user-images.githubusercontent.com/120374383/234536423-2660954c-3994-4967-a1c6-977bf5a374e5.png)
[...]

PR #1152, opened 2023-04-26
Intel MPX support

Nitr0-G:
Skipping all MPX instructions has been added in order to eliminate bugs caused on these Issues (https://github.com/avast/retdec/issues/1148 https://github.com/avast/retdec/issues/1135)

Proof that these bugs have been fixed, you can find in the same Issue(https://github.com/avast/retdec/issues/1148 https://github.com/avast/retdec/issues/1135)

Intel MPX is a dead technology that has not been supported by the Linux kernel since 2020(proof: https://www.phoronix.com/news/Intel-MPX-Is-Dead). It was only in the Skylake and Intel Goldmont(atom) architecture, consider all current processors do not support this technology. Zydis & capstone mistakenly disassembles instructions added to Intel MPX(Intel MPX adds 7 new instructions, as well as BND0-3 registers in x64 and x32 mode for more information, see here(https://intel-mpx.github.io/design/ )), a tool like Hiew also does not disassemble instructions of Intel MPX (https://fpic.in/VQ9yfJ1)

### DIFF
--- a/src/capstone2llvmir/capstone2llvmir_impl.cpp
+++ b/src/capstone2llvmir/capstone2llvmir_impl.cpp
@@ -237,10 +237,6 @@ Capstone2LlvmIrTranslator_impl<CInsn, CInsnOp>::translateOne(
 		llvm::IRBuilder<>& irb)
 {
 	TranslationResultOne res;
-	std::vector<unsigned char> MPX_INSTRS{
-	X86_INS_BNDMK, X86_INS_BNDCL, X86_INS_BNDCU, X86_INS_BNDMOV, X86_INS_BNDLDX, X86_INS_BNDSTX};
-	bool MpxInstr = false;
-	
 
 	// We want to keep all Capstone instructions -> alloc a new one each time.
 	cs_insn* insn = cs_malloc(_handle);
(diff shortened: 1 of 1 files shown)


PeterMatula:
Thanks for the investigation of the problem, however:

* Looks like the [tests](https://github.com/avast/retdec-regression-tests) are failing for you, see point below...
* This (i.e. `Capstone2LlvmIrTranslator_impl::translateOne()`) is not the right place where to add such code like you did. This method is meant to implement instruction translation on high-level. I.e. regardless the concrete instruction(s) (like `X86_INS_BND*` here) or architecture. By modifying the code here, you not only did not do it on the right place for X86, but you also affected translation of instructions for all the other architectures. Those `X86_INS_BND*` are Capstone's `x86_insn` enum items with an integer value. Their value may, and in fact it certainly does, conflict with values for other architectures like `arm_insn`. If you then do `if (insn[0].id == MpxIndexer)` you are gonna affect translation of all the other instructions - as instructions for all the architectures are handled by the `Capstone2LlvmIrTranslator_impl::translateOne()` method.
* Specific code for specific architectures comes to their modules like `src/capstone2llvmir/{arm, arm64, x86, ...}`. If you want to modify translation of x86 instruction, this is the general place to modify.
* Even there, your modification would not be the best way to do it. You don't want to make this kind of hackish fix on a method that processes all the instruction irrespective of their particular type. [Here](https://github.com/avast/retdec/blob/master/src/capstone2llvmir/x86/x86_init.cpp#L606) is the code that controls instruction translation - it assigns a translation routine to every instruction (looks like assignments for those `X86_INS_BND*` are missing, this is probably caused by migrating this code from older Capstone version which did have them to a newer version which does). Either a specific method is assigned to an instruction, or a method handling multiple instructions, or `translateNop()` if instruction is to be ignored, or `nullptr` which should trigger a best-effort automated translation.
* What probably happened here is the triggering of best-effort pseudo-instruction translation. And it failed, because it automatically tried to use registers, or whatever, which were not generated (again, probably because they were not in old Capstone version).
* So, if you want to generate nothing for `X86_INS_BND*`, simply add entries for these instructions [here](https://github.com/avast/retdec/blob/master/src/capstone2llvmir/x86/x86_init.cpp#L606) and map them to `&Capstone2LlvmIrTranslatorX86_impl::translateNop` e.g. as for `X86_INS_UD1`:
   ```cpp
   {X86_INS_UD1, &Capstone2LlvmIrTranslatorX86_impl::translateNop},
   ```
* While this might be an fine translation for these instructions, it would still be best to also implement register generation, if there are some registers in Capstone, which are not handled by `capstone2llvmir`. Register translation is handled by the same source code. The most important thing to fill is [this](https://github.com/avast/retdec/blob/master/src/capstone2llvmir/x86/x86_init.cpp#L92) map and they should ideally by generated and usable. However, x86 registers are sometimes tricky and a special handling needs to be implemented - it depends, have a look around that initialization source code to see what I mean.

I write all of this in case you want to properly fix your PR and understand the source code better. If you don't necessarily want that, let me know and I will do the fix myself. Thanks to your investigation of the problem it should be a relatively easy task for me :+1:

PR #1162, opened 2023-06-22
Fix some compilation errors

25077667:
- [fix: error: 'uint32_t' in namespace 'std' does not name a type;](https://github.com/avast/retdec/commit/8cd5ae166abbd9c2fd976a563ebb395ad43a431a) in include/retdec/utils/string.h

- [refactor: fix returning stack address in BaseErrors (https://github.com/avast/retdec/commit/730f800e301eba026c42a3a7c149a71ea32309f7) in include/retdec/capstone2llvmir/exceptions.h

Those error messages are evaluated on the stack (even though std::string would construct some char array on the heap), but it would be destructed when it left the stack frame.
I know that the LTO is default enabled in C++17, but it would still be destructed at the next stack frame.


25077667:
Apologies for the formatting issue caused by the IDE.

To improve the cleanliness of the working tree, I have removed commit [8cd5ae1](https://github.com/avast/retdec/pull/1162/commits/8cd5ae166abbd9c2fd976a563ebb395ad43a431a).

25077667:
There is another commit [7b28b20](https://github.com/avast/retdec/pull/1162/commits/7b28b2063f0a61da84d888292be81e655d477fa6) that also has a formatting issue, but it's a bit more complex to revise. If it's not a concern for you, I suggest keeping the changes made in this commit.

PR #1166, opened 2023-07-14
Fix .NET parser trying to read strings from String stream when out-of-bounds of the file

HoundThe:
When reading  strings from String stream of .NET binaries, there was no check, if the reading was actually succesfull (there is something to read in the file on that offset).

The string parsing is set to stop when it fails to read some string as they are read consecutively, so if offset X is out of bounds, X+1 should be too and there should be no reason to read further.


metthal (approved):
LGTM

PR #1167, opened 2023-07-15
MacOS workflow fix - Use new version of python and ubuntu that is not deprecated

HoundThe:
MacOS CI workflow is failing because python3.7 is deprecated and the build fails.

The doxygen build also fails due to deprecated github CI ubuntu version, it was changed to ubuntu-latest so it won't happen again.

### DIFF
--- a/.github/workflows/macOS/install-deps.sh
+++ b/.github/workflows/macOS/install-deps.sh
@@ -1,4 +1,4 @@
 #!/usr/bin/bash
 
-brew install pkg-config autoconf automake libtool openssl python@3.7
-brew link --overwrite python@3.7
+brew install pkg-config autoconf automake libtool openssl python@3.10
+brew link --overwrite python@3.10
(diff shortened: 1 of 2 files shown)


metthal (commented):
Doxygen seems to be still failing and I see some errors in the output. Is it just an old ubuntu or do we also need fixes in the documentation?

HoundThe:
> Doxygen seems to be still failing, and I see some errors in the output. Is it just an old Ubuntu, or do we also need fixes in the documentation?

I tried to build it locally, and it builds; I think it is just warnings, not actual real issues. From how the CI is set up, the test fails whenever it detects any non-empty doxygen build log.

```
          file="build/doc/doxygen/doxygen.log"
          echo "checking file: $file"
          content="$(cat $file)"
          if [ "$content" = "" ]; then
            echo "===> ok"
            else
            echo "===> fail:"
            echo "=========================================="
            echo "$content"
            echo "=========================================="
            exit 1
          fi
```
So either fix the warnings or make the checks less strict.

On the other hand, I've asked @PeterMatula about it and he said we should ignore it for now (the doxygen build has been ignored for quite a few commits already).

metthal (approved):

PR #1178, opened 2023-09-19
LLVM Error objects actually need to be consumed before their destruction

metthal:
I'm not sure how we haven't encountered this before but we actually need to consume errors to ignore them completely. Otherwise, they abort the program.


MatejKastak (approved):
LGTM :+1:

PR #1183, opened 2023-10-22
Fix all warnings in Apple Clang

takano32:
pick something good commits from network.

https://github.com/avast/retdec/network


takano32:
Sorry. I noticed this PR dupes with #1162

PR #1199, opened 2024-04-19
Adding support for MULX instruction.

lvella:
I am trying to implement the mulx instruction, but it is completely broken.

Whit this PR, the resulting c source from the file I am trying to decompile contains this broken typedef:
```
// ---------------- Integer Types Definitions -----------------

typedef int64_t int128_t;

```

The functions containing mulx are output as return 0, instead of whatever the code it has. E.g.:
```
// Address range: 0x1ce - 0x508
int64_t Fr_rawMMul(int64_t a1, int64_t a2, int64_t * a3) {
    // 0x4f0
    return 0;
}
```

This particular file I am trying this is the nasm output of https://github.com/powdr-labs/zkevm-prover/blob/26decaa2f6faa5367ca2eb24abfd2f23c0eb37fe/src/ffiasm/fr.asm

Can someone help me with it?


lvella:
Apparently this works for very simple examples, both using the last operand from register and from memory. But it breaks the case I actually care about.

